PR #12, opened 2018-05-08
[spec] Initial documentation of syntax

binji:
Includes the basic structure needed for the proposal, but no validation,
binary, text or execution.

### DIFF
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -241,7 +241,11 @@ Instructions in this group are concerned with linear :ref:`memory <syntax-mem>`.
      \K{i}\X{nn}\K{.}\STORE\K{16}~\memarg ~|~
      \K{i64.}\STORE\K{32}~\memarg \\&&|&
      \MEMORYSIZE \\&&|&
-     \MEMORYGROW \\
+     \MEMORYGROW \\&&|&
+     \MEMORYINIT~\dataidx \\&&|&
+     \MEMORYDROP~\dataidx \\&&|&
+     \MEMORYCOPY \\&&|&
+     \MEMORYFILL \\
    \end{array}
 
 Memory is accessed with |LOAD| and |STORE| instructions for the different :ref:`value types <syntax-valtype>`.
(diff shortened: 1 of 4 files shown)


binji:
I'm trying to push a copy to gh-pages but the scripts keep blowing up. :-/

binji:
BTW, I used `activeness` as the name of the type that chooses between `active` and `passive`, but I'm hoping someone can come up with something better! :-)

document/core/syntax/instructions.rst
```diff
@@ -260,12 +264,45 @@ The |MEMORYSIZE| instruction returns the current size of a memory.
 The |MEMORYGROW| instruction grows memory by a given delta and returns the previous size, or :math:`-1` if enough memory cannot be allocated.
 Both instructions operate in units of :ref:`page size <page-size>`.
 
+The |MEMORYINIT| instruction copies data from a :ref:`passive data segment <syntax-activeness>` into a memory.
+The |MEMORYDROP| instruction prevents further use of a passive data segment. This instruction is intended to be used as an optimization hint. After a data segment is dropped its data can no longer be retrieved, so the memory used by this segment may be freed.
+The |MEMORYCOPY| instruction copies data from a source memory region to a possibly overlapping destination region.
+The |MEMORYFILL| instruction sets all bytes in a region to a given byte.
```
rossberg: Nit: this sentence sounds a bit weird, in particular because "bytes" per se are just values, not mutable entities.

document/core/syntax/modules.rst
```diff
@@ -66,13 +72,17 @@ Each class of definition has its own *index space*, as distinguished by the foll
...
+   \production{element index} & \elemidx &::=& \u32 \\
+   \production{data index} & \dataidx &::=& \u32 \\
    \production{local index} & \localidx &::=& \u32 \\
    \production{label index} & \labelidx &::=& \u32 \\
    \end{array}
 
 The index space for :ref:`functions <syntax-func>`, :ref:`tables <syntax-table>`, :ref:`memories <syntax-mem>` and :ref:`globals <syntax-global>` includes respective :ref:`imports <syntax-import>` declared in the same module.
 The indices of these imports precede the indices of other definitions in the same index space.
 
+The index spaces for :ref:`element segments <syntax-elem>` and :ref:`data segments <syntax-data>` are relative to a given :ref:`table <syntax-table>` and :ref:`memory <syntax-mem>` respectively.
```
rossberg: That is strange. Why would passive segments be associated with specific tables or memories?

document/core/syntax/modules.rst
```diff
@@ -231,17 +241,19 @@ The |MELEM| component of a module defines a vector of *element segments* that in
 .. math::
    \begin{array}{llll}
    \production{element segment} & \elem &::=&
-     \{ \ETABLE~\tableidx, \EOFFSET~\expr, \EINIT~\vec(\funcidx) \} \\
+     \{ \ETABLE~\tableidx, \EOFFSET~\expr, \EINIT~\vec(\funcidx), \EACTIVENESS~\activeness \} \\
```
rossberg: I think a passive segment should have neither a table index nor an offset. So may be make these different records. Then you can leave the "activeness" flag implicit as well.

document/core/syntax/modules.rst
```diff
@@ -256,11 +268,13 @@ The |MDATA| component of a module defines a vector of *data segments* that initi
 .. math::
    \begin{array}{llll}
    \production{data segment} & \data &::=&
-     \{ \DMEM~\memidx, \DOFFSET~\expr, \DINIT~\vec(\byte) \} \\
+     \{ \DMEM~\memidx, \DOFFSET~\expr, \DINIT~\vec(\byte), \DACTIVENESS~\activeness \} \\
```
rossberg: Same here.

binji:
I've separated active and passive segments into their own records, PTAL!

document/core/syntax/modules.rst
```diff
@@ -236,12 +253,22 @@ The |MELEM| component of a module defines a vector of *element segments* that in
 
 The |EOFFSET| is given by a :ref:`constant <valid-constant>` :ref:`expression <syntax-expr>`.
 
+The |MPELEM| component of a module defines a vector of passive element segments. Each passive element segment only defines its contents.
```
rossberg: Ah, I thought putting them in the same section and index space, just with different records. Not sure I'd particularly fancy introducing new sections... Also, I could imagine there being future features that want to index both kinds uniformly.

binji: Oh, I see what you mean. I didn't think of these as being new sections, but I can see why that might be confusing.

binji:
OK, using the same module vectors now, w/ separate records.

document/core/syntax/modules.rst
```diff
@@ -239,29 +237,26 @@ starting with the smallest index not referencing a global :ref:`import <syntax-i
...
 
-The |MELEM| component of a module defines a vector of active element segments. Each active element segment defines the |ETABLE| and the starting |EOFFSET| in that table to initialize.
+The |MELEM| component of a module defines a vector of element segments. Each active element segment defines the |ETABLE| and the starting |EOFFSET| in that table to initialize. Each passive element segment only defines its contents.
 
 .. math::
    \begin{array}{llll}
    \production{element segment} & \elem &::=&
+     \EACTIVE~\aelem ~|~
+     \EPASSIVE~\pelem \\
+   \production{active element segment} & \aelem &::=&
```
rossberg: Nit: I don't think introducing auxiliary non-terminals for these is needed, just inline them above.

Also, both cases can be distinguished by structure, so the explicit active//passive tags aren't really necessary (cf. function instances, for example). One annoyance with such tags is that they prevent you from writing e.g. \elem.\einit and instead force you to always pattern-match the tag away first.

Otherwise looks good!

binji: I liked having the tags since it made it clearer which record is used for active vs. passive, but the text explains it so it's probably OK. Agreed it will be nicer to use `\einit` directly.

Not quite sure what you mean by function instances -- where are they distinguished by structure?

rossberg: They can either be regular or host functions, which are different records, see [here](http://webassembly.github.io/spec/core/exec/runtime.html#function-instances).

binji: Ah, right. I was looking at functions not function instances. :-)

binji: @rossberg BTW, I've just started to work on validation but I think I need to store whether the segment is active or passive in the index. I considered using a bool for this, but it seems better to use an enumeration. Any suggestions?

rossberg: Yes, I'd use the "activeness" enumeration you had in the abstract syntax in an earlier version. Don't have any good suggestion for a better name, though.

Btw, this makes me wonder: in principle, we could allow importing/exporting passive segments. Has that possibility ever been discussed?

PR #19, opened 2018-05-22
[spec] Document execution

binji:
Started to work on spec for execution. Still more work to do, but I thought it would be useful to get some early feedback.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -813,7 +813,7 @@ Table Instructions
 
 .. _exec-table.init:
 
-:math:`\TABLEINIT`
+:math:`\TABLEINIT~x`
 ..................
 
 1. Let :math:`F` be the :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>`.
@@ -826,7 +826,7 @@ Table Instructions
 
 5. Let :math:`\X{table}` be the :ref:`table instance <syntax-tableinst>` :math:`S.\STABLES[ta]`.
 
-6. Let :math:`tsz` be the length of :math:`\X{table}.\MIELEMS`.
+6. Let :math:`tsz` be the length of :math:`\X{table}.\TIELEM`.
 
 7. Assert: due to :ref:`validation <valid-table.init>`, :math:`F.\AMODULE.\MIELEMS[x]` exists.
 
@@ -872,7 +872,7 @@ Table Instructions
      \wedge & F.\AMODULE.\MIELEMS[x] = ea \\
      \wedge & S.\SELEM[ea] \neq \epsilon \\
      \wedge & (i + n \leq |S.\SELEM[ea].\EIINIT|) \\
-     \wedge & (j + n \leq |S.\STABLES[ta].\MIELEMS|) \\
+     \wedge & (j + n \leq |S.\STABLES[ta].\TIELEM|) \\
      \wedge & y^\ast = S.\SELEM[ea].\EIINIT[i \slice n]) \\
      \end{array}
    \\[1ex]
@@ -935,7 +935,7 @@ Table Instructions
 
 5. Let :math:`\X{table}` be the :ref:`table instance <syntax-tableinst>` :math:`S.\STABLES[ta]`.
 
-6. Let :math:`sz` be the length of :math:`\X{table}.\MIELEMS`.
+6. Let :math:`sz` be the length of :math:`\X{table}.\TIELEM`.
 
 7. Assert: due to :ref:`validation <valid-table.init>`, three values of :ref:`value type <syntax-valtype>` |I32| are on the top of the stack.
 
@@ -953,7 +953,7 @@ Table Instructions
 
     a. Trap.
 
-13. Let :math:`y^\ast` be the function address sequence :math:`\X{table}.\MIELEMS[i \slice n]`.
+13. Let :math:`y^\ast` be the function address sequence :math:`\X{table}.\TIELEM[i \slice n]`.
 
 14. :ref:`Initialize <initelem>` the table instance at address :math:`ta` starting from offset :math:`j` with the function address sequence :math:`y^\ast`.
 
@@ -966,9 +966,9 @@ Table Instructions
    \\ \qquad
      \begin{array}[j]{@{}r@{~}l@{}}
      (\iff & F.\AMODULE.\MITABLES[0] = ta \\
-     \wedge & (i + n \leq |S.\STABLES[ta].\MIELEMS|) \\
-     \wedge & (j + n \leq |S.\STABLES[ta].\MIELEMS|) \\
-     \wedge & y^\ast = S.\STABLES[ta].\MIELEMS[i \slice n]) \\
+     \wedge & (i + n \leq |S.\STABLES[ta].\TIELEM|) \\
+     \wedge & (j + n \leq |S.\STABLES[ta].\TIELEM|) \\
+     \wedge & y^\ast = S.\STABLES[ta].\TIELEM[i \slice n]) \\
      \end{array}
    \\[1ex]
    \begin{array}{lcl@{\qquad}l}
@@ -987,11 +987,9 @@ Table Instructions
 
 1. Let :math:`F` be the :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>`.
 
-2. For each :ref:`function index <syntax-funcidx>` :math:`x_i` of :math:`x^\ast`:
+2. For each :ref:`function address <syntax-funcaddr>` :math:`x_i` of :math:`x^\ast`:
 
-   a. Let :math:`\funcaddr_i` be the :ref:`function address <syntax-funcaddr>` :math:`F.\AMODULE.\MIFUNCS[x_i]`.
-
-   b. Replace :math:`S.\STABLES[\tableaddr].\TIELEM[o + i]` with :math:`\funcaddr_i`.
+   a. Replace :math:`S.\STABLES[\tableaddr].\TIELEM[o + i]` with :math:`x_i`.
 
 .. math::
    ~\\[-1ex]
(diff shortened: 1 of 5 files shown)


binji:
Rendered result: https://webassembly.github.io/bulk-memory-operations/core/bikeshed

binji:
@rossberg PTAL if you have time!

document/core/exec/instructions.rst
```diff
@@ -618,6 +618,76 @@ Memory Instructions
...
+
+   a. Trap.
+
+10. Let :math:`\X{data}` be the :ref:`data segment instance <syntax-datainst>` :math:`S.\SDATA[da]`.
+
+11. Assert: due to :ref:`validation <valid-memory.init>`, three values of :ref:`value type <syntax-valtype>` |I32| are on the top of the stack.
+
+12. Pop the value :math:`\I32.\CONST~s` from the stack.
+
+13. Pop the value :math:`\I32.\CONST~t` from the stack.
```
rossberg: Nit: the spec convention is to use `t` for value types, and i,j,k,m,n for integers, at least if they are single-letter. So either use those or somewhat longer names.

document/core/exec/instructions.rst
```diff
@@ -618,6 +618,76 @@ Memory Instructions
...
+
+16. If :math:`s + n > dsz`, then:
+
+    a. Trap.
+
+17. If :math:`t + n > msz`, then:
+
+    a. Trap.
+
+18. Let :math:`y` be the byte sequence :math:`\X{mem}.\DSIINIT[s \slice n]`.
```
rossberg: Nit: Use `b^\ast` instead of `y`.

document/core/exec/instructions.rst
```diff
@@ -618,6 +618,76 @@ Memory Instructions
...
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\iff & F.\AMODULE.\MIMEMS[0] = ma \\
+     \wedge & F.\AMODULE.\MIDATAS[x] = da \\
+     \wedge & (s + n \leq |S.\SDATA[da].\DSIINIT|) \\
+     \wedge & (t + n \leq |S.\SMEMS[ma].\MIDATA|) \\
+     \wedge & y = S.\SDATA[da].\DSIINIT[s \slice n]) \\
+     \end{array}
+   \\[1ex]
+   \begin{array}{lcl@{\qquad}l}
+   S; F; (\I32.\CONST~n)~(\I32.\CONST~t)~(\I32.\CONST~s)~(\MEMORYINIT~x) &\stepto& S; F; \TRAP
```
rossberg: This needs an "otherwise" as a side conditione, else it would always be allowed non-deterministically.

document/core/exec/modules.rst
```diff
@@ -371,6 +371,66 @@ New instances of :ref:`functions <syntax-funcinst>`, :ref:`tables <syntax-tablei
...
 
+.. index:: element, element instance, element address
+.. _alloc-elem:
+
+:ref:`Element segments <syntax-eleminst>`
+.........................................
+
+1. Let :math:`e` be the :ref:`element segment <syntax-elem>` to allocate.
+
+2. If :math:`e` is of the form :math:`\{ \EINIT~x^\ast \}`, then:
```
rossberg: What does this algorithm return otherwise? Should it be epsilon, i.e., the algo returns an optional address?

document/core/exec/modules.rst
```diff
@@ -371,6 +371,66 @@ New instances of :ref:`functions <syntax-funcinst>`, :ref:`tables <syntax-tablei
...
+   c. Let :math:`\funcaddr^\ast` be the concatenation of the function addresses :math:`\funcaddr_i`.
+
+   d. Let :math:`\eleminst` be the :ref:`element instance <syntax-eleminst>` :math:`\{ \ESIINIT~\funcaddr^\ast \}`.
+
+   e. Append :math:`\eleminst` to the |SELEM| of :math:`S`.
+
+.. math::
+   \begin{array}{rlll@{\qquad}l}
+   \allocelem(S, e, \moduleinst) &=& S', \elemaddr & (\iff e = \{ \EINIT~x^\ast \}) \\[1ex]
+   \allocelem(S, e, \moduleinst) &=& S & (\otherwise) \\[1ex]
```
rossberg: This only returns one thing, not a pair, which seems incoherent. I think it should be epsilon.

document/core/exec/modules.rst
```diff
@@ -371,6 +371,66 @@ New instances of :ref:`functions <syntax-funcinst>`, :ref:`tables <syntax-tablei
...
+   \mbox{where:} \hfill \\
```
rossberg: This should go one line above to the corresponding rule.

document/core/exec/modules.rst
```diff
@@ -371,6 +371,66 @@ New instances of :ref:`functions <syntax-funcinst>`, :ref:`tables <syntax-tablei
...
+
+.. index:: data, data instance, data address
+.. _alloc-data:
+
+:ref:`Data segments <syntax-datainst>`
+......................................
+
+1. Let :math:`d` be the :ref:`data segment <syntax-data>` to allocate.
+
+2. If :math:`d` is of the form :math:`\{ \DINIT~b^\ast \}`, then:
```
rossberg: Same here.

document/core/exec/modules.rst
```diff
@@ -439,23 +499,35 @@ and :math:`\val^\ast` the initialization :ref:`values <syntax-val>` of the modul
 
    a. Let :math:`\globaladdr_i` be the :ref:`global address <syntax-globaladdr>` resulting from :ref:`allocating <alloc-global>` :math:`\global_i.\GTYPE` with initializer value :math:`\val^\ast[i]`.
 
-6. Let :math:`\funcaddr^\ast` be the the concatenation of the :ref:`function addresses <syntax-funcaddr>` :math:`\funcaddr_i` in index order.
+6. For each :ref:`element segment <syntax-elem>` :math:`\elem_i` in :math:`\module.\MELEM`, do:
+
+   a. Let :math:`\elemaddr_i` be the :ref:`element address <syntax-elemaddr>` resulting from :ref:`allocating <alloc-elem>` :math:`\elem_i` for the :ref:`\module instance <syntax-moduleinst>` :math:`\moduleinst` defined below.
```
rossberg: Here you assume that the vector of elemaddr_i corresponds to elem_i, which is not the case because you only get addresses for passive elements. So I think what the module instance needs to store is not am elemaddr^* but an (elemaddr^?)^*. Otherwise all the indexing will be off, too.

document/core/exec/runtime.rst
```diff
@@ -158,6 +170,8 @@ and collects runtime representations of all entities that are imported, defined,
      \MITABLES & \tableaddr^\ast, \\
      \MIMEMS & \memaddr^\ast, \\
      \MIGLOBALS & \globaladdr^\ast, \\
+     \MIELEMS & \elemaddr^\ast, \\
```
rossberg: As mentioned above, I think these need to be optional addresses to represent indices of active segments correctly.

document/core/exec/runtime.rst
```diff
@@ -275,6 +289,42 @@ It holds an individual :ref:`value <syntax-val>` and a flag indicating whether i
...
+Element Instances
+~~~~~~~~~~~~~~~~~
+
+An *element instance* is the runtime representation of a :ref:`passive element segment <syntax-elem>`.
+It holds a vector of :ref:`function addresses <syntax-funcaddr>`.
+
+.. math::
+   \begin{array}{llll}
+   \production{(element instance)} & \eleminst &::=&
+     \{ \ESIINIT~\vec(\funcaddr) \} \\
```
rossberg: Nit: why not EIINIT?

document/core/exec/runtime.rst
```diff
@@ -275,6 +289,42 @@ It holds an individual :ref:`value <syntax-val>` and a flag indicating whether i
...
+Data Instances
+~~~~~~~~~~~~~~
+
+An *data instance* is the runtime representation of a :ref:`passive data segment <syntax-data>`.
+It holds a vector of :ref:`bytes <syntax-byte>`.
+
+.. math::
+   \begin{array}{llll}
+   \production{(data instance)} & \datainst &::=&
+     \{ \DSIINIT~\vec(\byte) \} \\
```
rossberg: Same Q here.

binji:
> One thing I was hoping for, though, is that with proper memory.init and table.init instructions we could get rid of the auxiliary init_data and init_elem ones, and replace their uses with the former. Is there a reason why that cannot work?

I think it's possible, and I started to do this, but it's pretty ugly unless we go with [option 2 here](https://github.com/WebAssembly/threads/issues/94#issuecomment-394643892).

Here's what I was thinking:

* Create element/data instances for all segments, even active ones
* `instantiate` executes something like `S;F; (i32.const |elem.data|) (i32.const 0) (elem.offset) (table.init x) (table.drop x)` for each active segment
* Similar for data segments

Without option 2, though, I guess you'd need to conditionally update the state only if none of the instructions trap. I started to write this, but wasn't quite sure how to define it, and thought we may need to change it for the threads proposal anyway.

rossberg:
Yes, I like that makes. Look slike the preferable route.

For option 1, wouldn't the only difference be that we keep the bounds checks as pre-checks in the instantiation itself?

document/core/exec/instructions.rst
```diff
@@ -690,6 +691,318 @@ Memory Instructions
...
+   ~\\[-1ex]
+   \begin{array}{l}
+   \begin{array}{lcl@{\qquad}l}
+   S; F; (\MEMORYDROP~x) &\stepto& S'; F; \epsilon
+   \end{array}
+   \\ \qquad
+     \begin{array}[j]{@{}r@{~}l@{}}
+     (\iff & F.\AMODULE.\MIMEMS[0] = ma \\
+     \wedge & F.\AMODULE.\MIDATAS[x] = da \\
+     \wedge & S' = S \with \SDATA[da] = \epsilon) \\
```
rossberg: Setting a SDATA slot to epsilon must allow for it in the definition of store, i.e., change the syntax to (\datainst^?)^*  (same for element segments).

binji:
> Sorry, dropped the ball on this review.

It's OK, I still haven't fixed the instantiation to use `MEMORYINIT`/`TABLEINIT` instead of `INITDATA`/`INITELEM`. Got a bit sidetracked with other stuff.

binji:
This is pretty out-of-date at this point, closing for now.

PR #28, opened 2018-09-06
Tweak segment encoding to keep memory/table index

AndrewScheidecker:
The current state seems to remove the placeholder table/memory index from the encoding to use it as an active flag.

This change proposes to keep the table/memory index by designating an index of -1 to mean "inactive segment", with 0+ indexing the module's memories/tables.

### DIFF
--- a/document/core/binary/modules.rst
+++ b/document/core/binary/modules.rst
@@ -330,6 +330,9 @@ It decodes into a vector of :ref:`element segments <syntax-elem>` that represent
    \production{element segment} & \Belem &::=&
      \hex{01}~~y^\ast{:}\Bvec(\Bfuncidx)
        &\Rightarrow& \{ \EINIT~y^\ast \} \\
+   \production{element segment} & \Belem &::=&
+     \hex{02}~~x{:}\Btableidx~~e{:}\Bexpr~~y^\ast{:}\Bvec(\Bfuncidx)
+       &\Rightarrow& \{ \ETABLE~x, \EOFFSET~e, \EINIT~y^\ast \} \\
    \end{array}
 
 .. note::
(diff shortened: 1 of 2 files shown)


AndrewScheidecker:
My goal is only to retain the memory/table index. Here's an alternative encoding that converts the memory index into a flag word, and adds an option to denote a non-zero memory index:
```
datasec ::= seg*:section_11(vec(data))           => seg
data    ::= 0x00 o:expr b*:vec(byte)             => {data memory 0,     offset o,     init b*, active true }
data    ::= 0x01 b*:vec(byte)                    => {data memory empty, offset empty, init b*, active false}
data    ::= 0x02 m:varuint32 o:expr b*:vec(byte) => {data memory m,     offset o,     init b*, active true }
data    ::= 0x03                                 => invalid
```

binji:
Something like your alternative encoding is what I was imagining we'd do when we want to add multiple memories. But as it is a validation error to have a non-zero index currently, I didn't think it was worthwhile to specify this yet.

AndrewScheidecker:
AFAICT, the MVP spec doesn't constrain the valid memory/table indices for data/elem segments. They're only constrained implicitly by the constraint on the number of memories/tables.

I'm thinking about how this extension combines with the reference-types extension, which allows more than one table. This extension redefining the interpretation of the table index to a flag byte means that it conflicts with that extension, where non-zero table indices are valid. A first byte of `0x01` for an encoded elem segment would be interpreted as "table index 1" in the reference-types extension, but as a passive segment encoding tag by this extension.

I think it's necessary to modify the reference-types extension to match this extension's encoding of elem segments. Assuming the reference-types extension wants the ability to encode elem segments for all tables, it makes sense to define an encoding that includes a table index. And while relaxing the constraint on number of memories doesn't seem imminent, it wouldn't hurt to keep encoding space for it in this extension.

AndrewScheidecker:
I updated the PR to the alternative format above (`0x02` -> segment with memory or table index), and also updated the spec document to reflect the proposed change.

binji:
OK, let's go ahead with this for now. Thanks!

rossberg:
I'm fine with this (sorry for being late).

PR #29, opened 2018-09-07
Add a second placeholder zero to the encoding of `memory.copy` and `table.copy`.

AndrewScheidecker:
This would make it simpler to extend those instructions to support multiple memories/tables, and copying between different memories/tables.

The current encoding has a single placeholder zero byte for those instructions, which allows extension to multiple memories/tables, but would require a more complicated encoding to add two immediate indices.


binji:
Hm, I'm not sure I'm convinced yet. We'll probably want a flag byte either way, so we can add other extensions later, so I wouldn't want to use these two bytes as indexes directly.

Issue #27, opened 2018-07-31
mem.init and mem.drop appear to break single-pass validation


julian-seward1:
Per my understanding, the new insns mem.init and mem.drop have an immediate (literal) operand, which is an index indicating the data segment from which the initialising data is to be copied.  I ran into a problem implementing validation for them, in that it's not possible to check that the literal value is in range at the time when validation passes over the code, because at that point the number of data segments isn't known -- the data segments appear in the binary format only after the code.

But I might misunderstand.  Can someone please clarify?

binji:
Yes, I see what you mean. Though you still can do it in one pass, you'd just have to remember the maximum data segment index and defer the validation error until you know how many data segments there are. I don't know of a good way around this, unfortunately.

AndrewScheidecker:
The init/drop ops are also verified to only use passive segments, so deferring the validation requires keeping track of all expected passive segment indices.

IMO it's better to just make an out-of-range or active segment a runtime trap instead of a validation error.

conrad-watt:
What was the original reason that init/drop cannot be used with active segments? Is there an expectation that implementations might want to optimise the representation of the module/instantiation process?

Our decision on this issue/PR (https://github.com/WebAssembly/threads/issues/94) / (https://github.com/WebAssembly/spec/pull/820) moves active segments towards being passive segments + a syntactic sugar for the start function. So maybe this could be revisited.

binji:
The original reason came from @lars-t-hansen in [this comment](https://github.com/WebAssembly/threads/issues/62#issuecomment-353282169) which I've inlined here:

> Just to reflect a conversation going on in #85: In my opinion, mem.init and mem.drop should only be usable with what's called "inactive" segments in that PR, ie, segments that are not copied into memory at instantiation time. If not, then mem.init will be sensitive to GC timing (an "active" segment referenced by mem.init might or might not have been removed by GC already because the module object was reaped) and instantiation will be sensitive to the timing of program execution, possibly in another thread (a concurrent mem.drop might have removed the segment). Either way we would need to spec failure semantics that we don't need if we simply say that "inactive" segments are for run-time use and "active" segments are for instantiation-time use. It should therefore be a verification failure for mem.init and mem.drop to reference segments that are not marked as "inactive".

conrad-watt:
[...]

PR #31, opened 2018-09-18
Validate segment index immediates regardless of passive-ness

AndrewScheidecker:
This is option 2a referenced here: https://github.com/WebAssembly/bulk-memory-operations/issues/27#issuecomment-421379736.

It allows segment indices that refer to active segments to pass validation, in anticipation of a future change to desugar active segments into init+drop instructions in the module's start function. That change implies that active segments will be dropped before any user code executes, and so will trap if referenced by a user init or drop instruction.

I've only changed the validation spec here, since the execution spec does not yet include this extension.

### DIFF
--- a/proposals/bulk-memory-operations/Overview.md
+++ b/proposals/bulk-memory-operations/Overview.md
@@ -180,12 +180,14 @@ the following new instructions:
 A passive segment has no initializer expression, since it will be specified
 as an operand to `memory.init` or `table.init`.
 
-Passive segments can also be discarded by using the following new instructions:
+Segments can also be discarded by using the following new instructions:
 
 * `memory.drop`: prevent further use of a data segment
 * `table.drop`: prevent further use of an element segment
 
-Attempting to drop an active segment is a validation error.
+An active segment is equivalent to a passive segment, but with an implicit
+`memory.init` followed by a `memory.drop` (or `table.init` followed by a
+`table.drop`) that is prepended to the module's start function.
 
 The data section is encoded as follows:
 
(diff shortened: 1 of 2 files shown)


proposals/bulk-memory-operations/Overview.md
```diff
@@ -180,12 +180,14 @@ the following new instructions:
...
 as an operand to `memory.init` or `table.init`.
 
-Passive segments can also be discarded by using the following new instructions:
+Segments can also be discarded by using the following new instructions:
 
 * `memory.drop`: prevent further use of a data segment
 * `table.drop`: prevent further use of an element segment
 
-Attempting to drop an active segment is a validation error.
+Active segments are equivalent to a passive segment, and an implicit `memory.init`
```
binji: This sentence sounds a bit odd to me, maybe "but with" instead of "and"?

proposals/bulk-memory-operations/Overview.md
```diff
@@ -211,10 +213,12 @@ step 11 of
 but it behaves as though the segment were specified with the source offset,
 target offset, and length as given by the `memory.init` operands.
 
-It is a validation error to use `memory.init` with an active segment.
+It is a validation error to use `memory.init` with an out-of-bounds segment index.
 
 A trap occurs if:
-* the segment is used after it has been dropped via `memory.drop`
+* the segment is used after it has been dropped via `memory.drop`. This includes
+  active memories that were dropped after being copied into memory during module
```
binji: s/memories/segments/

proposals/bulk-memory-operations/Overview.md
```diff
@@ -228,7 +232,10 @@ instruction. This instruction is intended to be used as an optimization hint to
 the WebAssembly implementation. After a memory segment is dropped its data can
 no longer be retrieved, so the memory used by this segment may be freed.
 
-It is a validation error to use `memory.drop` with an active segment.
+It is a validation error to use `memory.drop` with an out-of-bounds segment index.
+
+A trap occurs if the segment was already dropped. This includes active memories
```
binji: here too

AndrewScheidecker:
@binji Thanks, addressed your comments and updated the PR.

rossberg:
It seems like the passive/active distinction in a segtype is no longer inspected anywhere with this change, in which case we should simplify it to a single case.

PR #33, opened 2018-10-05
Add memory index for {memory,table}.init

binji:


binji:
I noticed this when reviewing https://github.com/WebAssembly/bulk-memory-operations/pull/32.

alexcrichton:
Is the memory index required for `memory.drop`? That's just dropping a data segment which otherwise isn't associated with a memory, right? (in that you can only drop a passive segment which doesn't have an associated memory index with it)

rossberg:
Agreeing with @alexcrichton that memory/table.drop don't need a memindex.

binji:
Oops, right :-)

Issue #49, opened 2019-01-25
Precise semantics of memory.copy


lars-t-hansen:
The overview currently says:

> Copy data from a source memory region to destination region; these regions may overlap: the copy is performed as if the source region was first copied to a temporary buffer, then the temporary buffer is copied to the destination region.

There's also some agreement about effectively having to copy low->high, and thus to write data up to the bounds limit for the target, see #10, so assume that.  This agreement seems to have been justified by the possible presence of an mprotect-like function that may make pages in the middle of a memory unavailable for reading and/or writing.

In unshared memory memory.copy is efficiently implementable without a buffer as follows, even in the presence of mprotect:

* if any source address is OOB or is unmapped then
  * trap
* if the tail of the target memory is OOB then
  * shrink the target memory down to the in-bounds range
  * set a flag that we must trap, the trap address is the first OOB address
* if any remaining target address is unmapped then
  * shrink the target memory down to the range that does not include the lowest unmapped address
  * set a flag that we must trap when we're done, the trap address will be the lowest unmapped address
* now copy src->target for the remaining target range, overlaps can be handled by copying high->low because that's not observable - there will be no faults
* if the trap flag was set then
  * trap

For shared memory it's different.  As some other agent can protect/unprotect pages of both source and target while the copying is happening, the copy may in fact have to use a temp buffer for all copies that can't be proven not to straddle any page boundaries.  (Suppose we don't, and just copy source->target, and a source page becomes unmapped and we must fault.  Now the target is partially changed, contrary to semantics.)  Clearly such a proof is possible at runtime, but even so.  Large copies that span pages will be forced to allocate a temp buffer equal to the size of the copy, I think, as the implementation may write no target addresses until all source addresses have been read.

Yet it's desirable to avoid such a temp buffer.  Though the requirement of a temp buffer won't affect the copy performance in unshared memories, it will put shared memory at a disadvantage, performance-wise.

Would it not be better to bow to reality here and say that if the copy is overlapping and src < dest, then copying is as-if high->low, and otherwise as-if low->high?  This is no less implementation-invariant and it allows a straightforward optimization to be used by implementations, and implementations that were going to do something else anyway ought not be worse off.  Implementations must still be a little careful around page boundaries but the buffer allocation is avoidable.

[...]

PR #50, opened 2019-01-25
Clarify trapping semantics

lars-t-hansen:
These are clarifications of the trapping semantics for memory.fill, memory.init, and memory.copy, and thus also for table.init and table.copy.  We clarify that there's a low->high write order and under what circumstances partially written data are observed.

This needs to be further elaborated once we agree on #49, but shouldn't be controversial at the moment.

### DIFF
--- a/proposals/bulk-memory-operations/Overview.md
+++ b/proposals/bulk-memory-operations/Overview.md
@@ -277,10 +277,14 @@ segment or target memory only occurs once the first byte that is
 outside the source or target is reached.  Bytes written before the
 trap stay written.
 
+(Data are read and written as-if individual bytes were read and
+written, but various optimizations are possible that avoid reading and
+writing only individual bytes.)
+
 Note that the semantics require bytewise accesses, so a trap that
 might result from, say, reading a sequence of several words before
 writing any, will have to be handled carefully: the reads that
-succeeded will have to be written, if possibe.
+succeeded will have to be written, if possible.
 
 ### `data.drop` instruction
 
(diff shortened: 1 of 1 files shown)


proposals/bulk-memory-operations/Overview.md
```diff
@@ -271,6 +271,17 @@ A trap occurs if:
...
+Initialization takes place bytewise from lower addresses toward higher
+addresses.  A trap resulting from an access outside the source data
+segment or target memory only occurs once the first byte that is
+outside the source or target is reached.  Bytes written before the
+trap stay written.
+
+Note that the semantics require bytewise accesses, so a trap that
+might result from, say, reading a sequence of several words before
+writing any, will have to be handled carefully: the reads that
+succeeded will have to be written, if possibe.
```
binji: sp: possible

proposals/bulk-memory-operations/Overview.md
```diff
@@ -271,6 +271,17 @@ A trap occurs if:
 Note that it is allowed to use `memory.init` on the same data segment more than
 once.
 
+Initialization takes place bytewise from lower addresses toward higher
+addresses.  A trap resulting from an access outside the source data
+segment or target memory only occurs once the first byte that is
+outside the source or target is reached.  Bytes written before the
+trap stay written.
+
+Note that the semantics require bytewise accesses, so a trap that
```
binji: I think I know what you mean here, but the text is still a little unclear. Perhaps mention that an implementation may want to optimize this by copying words instead of bytes?

lars-t-hansen: Added some prose.

lars-t-hansen:
I'm leaving the text for memory.copy as it is even though we've more or less agreed to remove the as-if intermediate buffer solution.  Separate PR.

PR #53, opened 2019-01-29
Clarify `*.init` can be used with active segments

alexcrichton:
On rereading the text of the proposal I *think* this is the intent.
Currently it reads as if `memory.init` and `table.init` can only be used
on passive segments, but the `DataCount` section I believe counts all
data segments, both active and passive, so when verifying function
bodies a segment index isn't known whether to be active or passive.

### DIFF
--- a/proposals/bulk-memory-operations/Overview.md
+++ b/proposals/bulk-memory-operations/Overview.md
@@ -243,8 +243,9 @@ TODO: coordinate with other proposals to determine the binary encoding for `ref.
 
 ### `memory.init` instruction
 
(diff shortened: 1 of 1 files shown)


binji:
Right, active and passive segments share the same index space, but the `*.init` instructions only work with passive segments, and trap if they are used with active segments. This is because active segments behave as though they are passive segments that have been dropped during module initialization.

alexcrichton:
I seem to have suffered again from "didn't read the next few sentences closely enough again", my mistake!

Thanks for the update!

PR #54, opened 2019-02-01
[interpreter] Implement bulk memory operations

binji:
This adds the new operations, as well as the new data and element
segment formats.

Implemented:

* Text parsing/writing
* Binary encoding/decoding
* Validation

TODO:

* Tests
* DataCount section
* Proper encoding of passive element segments (using ref.null etc)
* Evaluation

### DIFF
--- a/interpreter/binary/decode.ml
+++ b/interpreter/binary/decode.ml
@@ -607,23 +607,23 @@ let code_section s =
 (* Element section *)
 
 let segment dat s =
-  let flag = u8 s in
-  match flag with
-  | 0x00 ->
+  match vu32 s with
+  | 0l ->
+    let index = Source.(0l @@ Source.no_region) in
     let offset = const s in
-    let stype = Active {index = Source.((0l) @@ Source.no_region); offset} in
+    let desc = Active {index; offset} in
     let init = dat s in
-    {stype; init}
-  | 0x01 ->
-    let stype = Passive in
+    {desc; init}
+  | 1l ->
+    let desc = Passive in
     let init = dat s in
-    {stype; init}
-  | 0x02 ->
+    {desc; init}
+  | 2l ->
     let index = at var s in
     let offset = const s in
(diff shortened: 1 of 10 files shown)


interpreter/text/parser.mly
```diff
@@ -730,13 +772,13 @@ module_fields1 :
       { m with funcs = funcs @ m.funcs;
         imports = ims @ m.imports; exports = exs @ m.exports } }
   | elem module_fields
-    { fun c -> let mf = $2 c in
-      fun () -> let m = mf () in
-      {m with elems = $1 c :: m.elems} }
+    { fun c -> let ef = $1 c in let mf = $2 c in
```
rossberg: Guessing you had a lot of fun figuring out the right phasing here. :) Sorry for the mess.

interpreter/binary/decode.ml
```diff
@@ -589,10 +607,24 @@ let code_section s =
 (* Element section *)
 
 let segment dat s =
-  let index = at var s in
```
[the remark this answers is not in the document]
binji: In the overview this is a `vu32`, fixed it here so it's not a breaking change.

interpreter/valid/valid.ml
```diff
@@ -29,6 +27,7 @@ type context =
   results : value_type list;
   labels : stack_type list;
 }
+and segment_type = Seg
```
rossberg: Move this to types.ml?

interpreter/syntax/ast.ml
```diff
@@ -147,7 +147,7 @@ and memory' =
 type 'data segment = 'data segment' Source.phrase
 and 'data segment' =
 {
-  stype : segment_type;
+  desc : segment_desc;
```
rossberg: Since there are other xdesc's in the AST already, I'd call this `sdesc`.

PR #59, opened 2019-02-20
[Interpreter] Decode/encode the data count section

binji:


interpreter/binary/decode.ml
```diff
@@ -679,13 +686,18 @@ let module_ s =
...
+  iterate custom_section s;
   let func_bodies = code_section s in
   iterate custom_section s;
   let data = data_section s in
   iterate custom_section s;
   require (pos s = len s) s (len s) "junk after last section";
   require (List.length func_types = List.length func_bodies)
     s (len s) "function and code section have inconsistent lengths";
+  require (data_count = None ||
+           data_count = Some (Int32.of_int (List.length data)))
```
rossberg: Style nit: avoid indentation that breaks with non-proportional fonts

PR #60, opened 2019-02-21
[Interpreter] Implement memory.copy, memory.fill

binji:


interpreter/runtime/memory.ml
```diff
@@ -143,3 +144,30 @@ let store_packed sz mem a o v =
...
+
+let check_bounds mem a = if I64.ge_u a (bound mem) then raise Bounds
+
+let fill mem a v n =
+  let rec loop a n =
+    if n = 0l then () else begin
+      store_byte mem a v;
+      loop (Int64.add a 1L) (Int32.sub n 1l)
+    end
+  in if n = 0l then check_bounds mem a
```
rossberg: I think you can avoid the condition and do `check_bounds mem (a+n)` unconditionally after the loop. Similarly for copy.

interpreter/runtime/memory.ml
```diff
@@ -143,3 +144,30 @@ let store_packed sz mem a o v =
     | I64 x -> x
     | _ -> raise Type
   in storen mem a o n x
+
+let check_bounds mem a = if I64.ge_u a (bound mem) then raise Bounds
+
+let fill mem a v n =
+  let rec loop a n =
+    if n = 0l then () else begin
```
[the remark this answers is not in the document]
binji: I couldn't see how to use a for loop since it wanted an int not an int32.

interpreter/runtime/memory.ml
binji: I don't think that works -- it's OK to end a region outside of bounds, but a zero-length region still fails even though it doesn't write any data.

binji: Oh, sorry I see what you mean -- after n is reduced to 0.

binji: Ah, but you can do it unconditionally at the beginning instead.

binji: Sorry, I finally see what you mean -- we currently allow the following in the MVP:

```wasm
(module
  (memory 1)
  (data (i32.const 0x1_0000) "")
)
```

So init/copy/fill should behave the same.

PR #62, opened 2019-02-22
[Interpreter] Initialized segments persist after instantiation failure

binji:
This behavior is a breaking change from the v1 spec, but it is unlikely
that any programs depend on it.


binji:
Hm, not sure how to change a PR to squash to a different branch.

PR #65, opened 2019-02-27
[Interpreter] New syntax for passive elem segments

binji:
It uses instruction encoding instead of assuming that each element is a
function variable:

```
0xd0 -> ref.null
0xd2 x -> ref.func x
```


binji:
@rossberg Landing this one too, I'll address feedback in a future PR.

PR #66, opened 2019-03-01
[Interpreter] Fix bugs found by Lars' tests

binji:
* The data count section is required if the `memory.init` or `data.drop`
  instructions are used.

* `memory.init` had an off-by-one bug when reading at the end of a segment.

* All instructions that operate on regions need to compare length to 0
  using unsigned comparison (e.g. `I32.gt_u n 0l`).

* Converting length from 32- to 64-bit now uses
  `I64_convert.extend_i32_u`, instead of `Int64.of_int32`, since the
  latter will sign-extend.

* The `table.copy` overlap test now uses an unsigned comparison.

* The passive element text syntax still allows function indexes, e.g.
  `(elem passive $f1 $f2)`.

### DIFF
--- a/interpreter/text/parser.mly
+++ b/interpreter/text/parser.mly
@@ -584,25 +584,28 @@ active_elemref_list :
       fun c lookup -> List.map f ($1 c lookup) }
 
 elem :
-  | LPAR ELEM bind_var_opt PASSIVE passive_elemref_list RPAR
+  | LPAR ELEM bind_var_opt PASSIVE elem_type passive_elemref_list RPAR
     { let at = at () in
       fun c -> ignore ($3 c anon_elem bind_elem);
-      fun () -> Passive ($5 c) @@ at }
+      fun () -> Passive ($5, ($6 c)) @@ at }
   | LPAR ELEM bind_var var offset active_elemref_list RPAR
     { let at = at () in
       fun c -> ignore (bind_elem c $3);
       fun () ->
-      Active {index = $4 c table; offset = $5 c; init = $6 c func} @@ at }
+      let init = FuncRefType, ($6 c func) in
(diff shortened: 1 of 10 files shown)


interpreter/text/parser.mly
```diff
@@ -583,25 +584,28 @@ active_elemref_list :
       fun c lookup -> List.map f ($1 c lookup) }
 
 elem :
-  | LPAR ELEM bind_var_opt PASSIVE passive_elemref_list RPAR
+  | LPAR ELEM bind_var_opt PASSIVE elem_type passive_elemref_list RPAR
```
rossberg: Hm, so with the elem_type being mandatory, it seems like we could drop the extra "passive" keyword?

interpreter/text/parser.mly
```diff
@@ -572,6 +572,7 @@ offset :
 elemref :
   | LPAR REF_NULL RPAR { let at = at () in fun c -> Null @@ at }
   | LPAR REF_FUNC var RPAR { let at = at () in fun c -> Func ($3 c func) @@ at }
+  | var { let at = at () in fun c -> Func ($1 c func) @@ at }
```
rossberg: I would strongly prefer dropping this form, as it is an odd one out that doesn't really scale beyond what we currently have.

interpreter/text/parser.mly
binji: Yes, I think you're right.

interpreter/text/parser.mly
binji: That's true, but the other forms are quite a bit more verbose in the examples. Still, I'm fine dropping it.

rossberg: Okay, please see PR #84.

PR #67, opened 2019-03-05
Bulk memory operations test cases

lars-t-hansen:
These were translated from the Firefox test cases, with some adjustments (Firefox has some too-strict bounds checking bugs).  All pass in the reference interpreter with PRs 64, 65, and 66 applied.


test/meta/Makefile
```diff
@@ -0,0 +1,27 @@
+# SpiderMonkey shell
+JSSHELL=~/m-i/js/src/build-debug/dist/bin/js -e 'const WITH_SHARED_MEMORY=false;' -f common.js
```
binji: We might want to add a way to use `d8` or `node` here too, but this is fine for now.

test/meta/generate_table_copy.js
```diff
@@ -0,0 +1,319 @@
...
+// - arithmetic overflow on target addresses
+//
+// for each of those,
+//
+// - src address oob
+// - target address oob
+// - both oob
+//
+// Note we do not test the multi-table case here because that is part of the
+// reftypes proposal; tests are in the gc/ subdirectory.
```
binji: we should remove

PR #70, opened 2019-03-06
Remove irrelevant comment + support node.js

lars-t-hansen:


test/meta/Makefile
```diff
@@ -1,5 +1,10 @@
+SHARED_MEM=false
+
 # SpiderMonkey shell
-JSSHELL=~/m-i/js/src/build-debug/dist/bin/js -e 'const WITH_SHARED_MEMORY=false;' -f common.js
+JSSHELL=~/m-i/js/src/build-debug/dist/bin/js -e 'const WITH_SHARED_MEMORY=$(SHARED_MEM);' -f common.js
+
+# Node.js
+#JSSHELL=./noderun.sh $(SHARED_MEM)
```
binji: yikes, it's unfortunate that node.js doesn't have a nice way to do this. (maybe `fs.readFileSync` and `eval`?) I was using d8, where it's just `d8 -e "..." common.js`.

lars-t-hansen: Patches accepted :-)  The indirection through the shell is indeed unfortunate but at least it allows people to use node, which is everywhere.

PR #71, opened 2019-03-08
[interpreter] Refactor element type in passive segments

rossberg:
* Move a segment's element type into a separate record field for passive segments (only), avoiding dummy appearances in active segments.

* Fix a minor bug in encoding condition for data count section, which caused one test to fail.

* Check all global decoding side conditions in the same place.

* Minor naming consolidation.

### DIFF
--- a/interpreter/binary/decode.ml
+++ b/interpreter/binary/decode.ml
@@ -5,12 +5,12 @@ type stream =
   name : string;
   bytes : string;
   pos : int ref;
-  has_data_count : bool ref;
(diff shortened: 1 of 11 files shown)


rossberg:
I also tried to get rid of piggy-backing the decoder's stream abstraction to pass around the data count flag. That is rather hacky and does not translate well to an actual spec. But as you probably observed as well the alternatives are intrusive. I suppose that is a consequence of the fact that the nature of this condition is somewhat outside mere decoding, halfway to validation.

PR #72, opened 2019-03-08
Replace decoding flag with predicate

rossberg:
Closer to a declarative spec. Symmetrically, enables to only encode data count section when needed.

(Btw, I noticed that the data count section is still missing in the spec.)


binji:
> (Btw, I noticed that the data count section is still missing in the spec.)

Yeah, I haven't spent much time on the spec text in a while.

Issue #74, opened 2019-03-08
Spec still uses old names for data/elem.drop instructions


rossberg:

binji:
Fixed in #75, and one missing case in fcef8d6dd27b1a384c9e4c1e45bb9cdae5c3d133.

PR #75, opened 2019-03-30
[spec/interpreter] Rename elem/data.drop; update README; reorder

rossberg:
Update the interpreter README with the added instructions. While at it, also update names of .drop operators in spec (addressing #74). Moreover, reorder/regroup instructions/spec sections so that they will fit more nicely when combined with reftypes' table instructions; the order of the complete set would be:
```
local.{get,set,tee}
global.{get,set}
table.{get,set}
table.{size,grow,fill,copy,init}
elem.drop
load,store
memory.{size.grow,fill,copy,init}
data.drop
```


document/core/valid/instructions.rst
```diff
@@ -480,83 +520,43 @@ Memory Instructions
...
-.. _valid-table.copy:
+:math:`\DATADROP~x`
+...................
 
-:math:`\TABLECOPY`
-.....................
+* The data segment :math:`C.\CDATA[x]` must be defined in the context.
 
-* The table :math:`C.\CTABLES[0]` must be defined in the context.
+* The :ref:`segment type <syntax-segtype>` :math:`C.\CDATA[x]` must be |SPASSIVE|.
```
binji: This isn't correct anymore; not sure if you want to remove in this PR, since it's mostly just moving things around.

interpreter/exec/eval.ml
```diff
@@ -279,33 +266,46 @@ let rec step (c : config) : config =
...
         with exn -> vs', [Trapping (memory_error e.at exn) @@ e.at])
 
-      | TableInit x, I32 n :: I32 s :: I32 d :: vs' ->
-        let tab = table frame.inst (0l @@ e.at) in
-        (match !(elem frame.inst x) with
-        | Some es ->
-          (try Table.init tab es d s n; vs', []
-          with exn -> vs', [Trapping (table_error e.at exn) @@ e.at])
-        | None -> vs', [Trapping "element segment dropped" @@ e.at]
+      | MemoryInit x, I32 n :: I32 s :: I32 d :: vs' ->
```
binji: Move `MemoryFill` before `MemoryCopy`?

document/core/valid/instructions.rst
rossberg: Right, it's probably better to separate such fixes. I'll create an issue to track.

interpreter/exec/eval.ml
rossberg: Done.

PR #81, opened 2019-04-05
[spec] Update instruction index

binji:
Fixes issue #77.


binji:
The execution links are dead, but I'll add those next.

PR #82, opened 2019-04-06
[spec] Proper binary format for element segments

binji:
The element segment binary format allows for `ref.null` as well as
`ref.func`.

TODO: text format

### DIFF
--- a/document/core/binary/modules.rst
+++ b/document/core/binary/modules.rst
@@ -334,8 +334,8 @@ It decodes into a vector of :ref:`element segments <syntax-elem>` that represent
      \hex{02}~~x{:}\Btableidx~~e{:}\Bexpr~~y^\ast{:}\Bvec(\Bfuncidx)
        &\Rightarrow& \{ \ETABLE~x, \EOFFSET~e, \EINIT~((\REFFUNC~y)~\END)^\ast \} \\
    \production{elemexpr} & \Belemexpr &::=&
-     \hex{D0} &\Rightarrow& \REFNULL~\END \\ &&|&
-     \hex{D2}~x{:}\Bfuncidx &\Rightarrow& (\REFFUNC~x)~\END \\
+     \hex{D0}~\hex{0B} &\Rightarrow& \REFNULL~\END \\ &&|&
+     \hex{D2}~x{:}\Bfuncidx~\hex{0B} &\Rightarrow& (\REFFUNC~x)~\END \\
    \end{array}
 
 .. note::
--- a/document/core/binary/modules.rst
+++ b/document/core/binary/modules.rst
@@ -314,6 +314,7 @@ It decodes into an optional :ref:`start function <syntax-start>` that represents
    single: element; segment
 .. _binary-elem:
 .. _binary-elemsec:
+.. _binary-element:
 
 Element Section
 ~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


document/core/binary/modules.rst
```diff
@@ -327,13 +328,16 @@ It decodes into a vector of :ref:`element segments <syntax-elem>` that represent
...
+       &\Rightarrow& \{ \ETABLE~0, \EOFFSET~e, \EINIT~(\EREFFUNC~y)^\ast \} \\
    \production{element segment} & \Belem &::=&
-     \hex{01}~~y^\ast{:}\Bvec(\Bfuncidx)
+     \hex{01}~~y^\ast{:}\Bvec(\Belement)
        &\Rightarrow& \{ \EINIT~y^\ast \} \\
    \production{element segment} & \Belem &::=&
      \hex{02}~~x{:}\Btableidx~~e{:}\Bexpr~~y^\ast{:}\Bvec(\Bfuncidx)
-       &\Rightarrow& \{ \ETABLE~x, \EOFFSET~e, \EINIT~y^\ast \} \\
+       &\Rightarrow& \{ \ETABLE~x, \EOFFSET~e, \EINIT~(\EREFFUNC~y)^\ast \} \\
+   \production{element} & \Belement &::=&
```
binji: Better name for this? currently we use `elem` to mean "element segment", and I'm not sure it's worth changing that everywhere.

document/core/util/macros.def
```diff
@@ -254,6 +254,9 @@
 .. |EOFFSET| mathdef:: \xref{syntax/modules}{syntax-elem}{\K{offset}}
 .. |EINIT| mathdef:: \xref{syntax/modules}{syntax-elem}{\K{init}}
 
+.. |EREFNULL| mathdef:: \xref{syntax/modules}{syntax-element}{\K{ref.null}}
+.. |EREFFUNC| mathdef:: \xref{syntax/modules}{syntax-element}{\K{ref.func}}
```
rossberg: I think it's fine to already call these macros REFNULL and REFFUNC, as in the reftypes proposal.

document/core/valid/modules.rst
```diff
@@ -161,7 +161,10 @@ Element segments :math:`\elem` are classified by :ref:`segment types <syntax-seg
 * The expression :math:`\expr` must be :ref:`constant <valid-constant>`.
 
 * For each :math:`y_i` in :math:`y^\ast`,
```
rossberg: This meta variable should change to `e` as well.

document/core/valid/modules.rst
```diff
@@ -161,7 +161,10 @@ Element segments :math:`\elem` are classified by :ref:`segment types <syntax-seg
...
-  the function :math:`C.\CFUNCS[y]` must be defined in the context.
+
+  * :math:`y_i` must be :math:`\EREFFUNC~z`.
```
rossberg: Can you factor this out into a separate rule for elemexpr that is shared between active/passive segment types?

I don't think you need to check here that it is a ref.func for an active segment, since that's already enforced by the binary/text format. At least in the reference interpreter we don't make that distinction during validation.

document/core/valid/modules.rst
```diff
@@ -174,7 +177,7 @@ Element segments :math:`\elem` are classified by :ref:`segment types <syntax-seg
      \qquad
      C \vdashexprconst \expr \const
      \qquad
-     (C.\CFUNCS[y] = \functype)^\ast
+     (y = \EREFFUNC~z \wedge C.\CFUNCS[z] = \functype)^\ast
```
rossberg: Nit: Use meta variables e and y instead of y and z.

document/core/valid/modules.rst
```diff
@@ -184,14 +187,17 @@ Element segments :math:`\elem` are classified by :ref:`segment types <syntax-seg
 ......................................................
 
 * For each :math:`y_i` in :math:`y^\ast`,
```
rossberg: Nit: Change metavariable to e.

document/core/binary/modules.rst
binji: I used `elemexpr`. Rebasing seems like it might be more confusing, to me unfortunately...

document/core/binary/modules.rst
```diff
@@ -327,13 +328,16 @@ It decodes into a vector of :ref:`element segments <syntax-elem>` that represent
      \X{seg}^\ast{:}\Bsection_9(\Bvec(\Belem)) &\Rightarrow& \X{seg} \\
    \production{element segment} & \Belem &::=&
      \hex{00}~~e{:}\Bexpr~~y^\ast{:}\Bvec(\Bfuncidx)
-       &\Rightarrow& \{ \ETABLE~0, \EOFFSET~e, \EINIT~y^\ast \} \\
+       &\Rightarrow& \{ \ETABLE~0, \EOFFSET~e, \EINIT~(\EREFFUNC~y)^\ast \} \\
    \production{element segment} & \Belem &::=&
```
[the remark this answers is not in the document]
binji: You mean like this? `&&|&`?

document/core/binary/modules.rst
```diff
@@ -327,13 +328,14 @@ It decodes into a vector of :ref:`element segments <syntax-elem>` that represent
...
-   \production{element segment} & \Belem &::=&
+       &\Rightarrow& \{ \ETABLE~0, \EOFFSET~e, \EINIT~((\REFFUNC~y)~\END)^\ast \} \\ &&|&
+     \hex{01}~~e^\ast{:}\Bvec(\Belemexpr)
+       &\Rightarrow& \{ \EINIT~e^\ast \} \\ &&|&
      \hex{02}~~x{:}\Btableidx~~e{:}\Bexpr~~y^\ast{:}\Bvec(\Bfuncidx)
-       &\Rightarrow& \{ \ETABLE~x, \EOFFSET~e, \EINIT~y^\ast \} \\
+       &\Rightarrow& \{ \ETABLE~x, \EOFFSET~e, \EINIT~((\REFFUNC~y)~\END)^\ast \} \\
+   \production{elemexpr} & \Belemexpr &::=&
+     \hex{D0} &\Rightarrow& \REFNULL~\END \\ &&|&
+     \hex{D2}~x{:}\Bfuncidx &\Rightarrow& (\REFFUNC~x)~\END \\
```
rossberg: Ah, I was rather referring to the end opcode (`\hex{0B}`) missing on the lhs. We don't need to represent it in the abstract syntax, but if you did you'd also need to adjust syntax/modules.rst.

binji: Oh, right. Well `\expr` uses `\END` in the syntax, so we may as well do that here until we merge this with the reference types proposal.

PR #83, opened 2019-04-15
[spec] Simplify datacount side condition

rossberg:
This makes the side condition more generic by expressing it in terms of occurrences of data indices instead of specific instructions. It also simplifies slightly by using just one m.

### DIFF
--- a/document/core/binary/modules.rst
+++ b/document/core/binary/modules.rst
@@ -529,7 +529,7 @@ Furthermore, it must be present if any :math:`data index <syntax-dataidx>` occur
        \MIMPORTS~\import^\ast, \\
        \MEXPORTS~\export^\ast ~\} \\
        \end{array} \\ &&&
-     (\iff m^? \neq \epsilon \vee \F{dataidx}(\X{code}^n) = \emptyset) \\
+     (\iff m^? \neq \epsilon \vee \freedataidx(\X{code}^n) = \emptyset) \\
    \end{array}
 
 where for each :math:`t_i^\ast, e_i` in :math:`\X{code}^n`,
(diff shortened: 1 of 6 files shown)


document/core/binary/modules.rst
```diff
@@ -512,7 +514,7 @@ The :math:`\MEMORYINIT` and :math:`\DATADROP` instructions can only be used if t
      \Bcustomsec^\ast \\ &&&
      \X{code}^n{:\,}\Bcodesec \\ &&&
      \Bcustomsec^\ast \\ &&&
-     \data^{\X{m'}}{:\,}\Bdatasec \\ &&&
+     \data^m{:\,}\Bdatasec \\ &&&
```
binji: Is there syntax for this elsewhere? It seems odd to me to take `m^?` and remove the `^?` -- I suppose I'd guess that it would produce `0`, but it would be nice to mention that somewhere.

document/core/binary/modules.rst
```diff
@@ -522,30 +524,19 @@ The :math:`\MEMORYINIT` and :math:`\DATADROP` instructions can only be used if t
...
        \MSTART~\start^?, \\
        \MIMPORTS~\import^\ast, \\
        \MEXPORTS~\export^\ast ~\} \\
-      \end{array} \\ &&&
-     (\begin{array}[t]{@{}c@{~}l@{}}
-      \iff & m^? \neq \epsilon \\
-      \vee & \forall (t^\ast, e) \in \X{code}^n, \MEMORYINIT \notin e \wedge \DATADROP \notin e) \\
-      \end{array} \\
+       \end{array} \\ &&&
+     (\iff m^? \neq \epsilon \vee \F{dataidx}(\X{code}^n) = \emptyset) \\
```
binji: Can we link this `\F` set to the definition below?

document/core/binary/modules.rst
rossberg: I seem to remember that I've already used this kind of trick elsewhere, but I cannot find it right now. I added an additional bullet to make this kind of thing explicit in the conventions.

document/core/binary/modules.rst
rossberg: Done.

PR #84, opened 2019-04-16
Slight simplifications to text format

rossberg:
As previously suggested:

* Drop `passive` keyword (no longer needed for disambiguation).
* Remove special case syntax for passive funcref segments (would not work for other element types).


binji:
@lars-t-hansen for changes to the text format

I'm OK w/ this change in general, though I'm curious your thoughts about the generated tests here. Personally, I'd prefer either making the change in the JS source, or removing the generators entirely. Since the tests are so verbose, I'm hesitant to suggest we modify them by hand from now on.

rossberg:
Good point. Personally, I am not convinced it's worthwhile to maintain a dependency on the entire node ecosystem for a handful of generated tests. Also, it's more difficult to add simple tests that way.

lars-t-hansen:
Yeah, no objections, except SpiderMonkey's syntax is drifting further and further away from the official syntax... :-/

rossberg:
Should I remove the generators then (probably incorporating some of the comments into the tests)?

rossberg:
Friendly ping.

binji:
Yeah, I'm OK w/ removing the generators. We already have tests that are like this (e.g. `f32.wast`, `skip-stack-guard-page.wast`) and those don't seem to have been too much of an issue.

rossberg:
Okay, done.

binji:
Oops, just noticed PR #86. I'll try to merge these two.

PR #88, opened 2019-05-15
[spec] Add memory.fill to exec section

binji:


binji:
I started on `memory.copy` too, but I couldn't find a way to do it recursively without an administrative instruction. The direction of the copy needs to be preserved throughout, but the inductive case may remove the overlap. Any ideas?

rossberg:
FWIW, if there is a chance to still tweak this semantics then I'd vote that we do. As mentioned, it should only affect racy, i.e., broken, programs.

lars-t-hansen:
> As for copy, I was under the impression that we always copy in reverse direction when the source offset is larger than the destination offset. But upon rereading the design doc, I realise that is not the case.
> 
> Is there a reason to apply the more complicated condition that depends on overlap instead of only comparing offsets? Since the direction in the non-overlapping case doesn't matter (unless a program is racy anyway), why not use the simplest (and presumably cheapest) possible condition?

I don't know about "reason" but if we copy into increasing addresses when src < dest but there's no overlap, then the OOB behavior (on the first write) is different than if we copy into decreasing addresses.  It's hard to believe that a change here would have much practical impact for the programmer, but for the compiler the simpler test is appealing.

I also don't know if there might be hardware considerations that make copying into increasing addresses preferable (eg specific block move instructions), but our current semantics make that copy direction more prevalent.

(The REP MOVS family on x86 can move upward or downward according to the setting of DF.  The Intel optimization manual does not indicate that the direction has a performance impact.)

rossberg:
@lars-t-hansen, ah, good point about OOB traps. But I agree that this should not matter, it being a fatal error condition, which programs are ill-advised to depend on.

binji:
It seems that the [glibc C](https://sourceware.org/git/?p=glibc.git;a=blob;f=string/memmove.c;h=84b4339ecb1edeebf7dc5d2884b57809fef34ac6;hb=HEAD), [musl](http://git.musl-libc.org/cgit/musl/tree/src/string/memmove.c), and [newlib](https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=newlib/libc/string/memmove.c;h=da5dfdbdd68a97a23863945709195d70e98c3267;hb=refs/heads/master) implementations of `memmove` use our current semantics (reverse copy only on overlap). Though it looks like the [LLVM lowering pass](http://llvm.org/doxygen/LowerMemIntrinsics_8cpp_source.html#l00265), and [freebsd libc](https://github.com/freebsd/freebsd/blob/master/lib/libc/string/bcopy.c#L87) performs the simpler check.

The glibc x64 implementation uses many techniques, but does seem to [check for overlapping regions](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S;h=2e9d86bd33125a0c8df14304e1fe060efda2952d;hb=HEAD#l253) when the size is large, and never uses `rep movsb` backward unless `VEC_SIZE == 16` (I think ... it's a bit hard to understand).

Anyway, I think we could change to a simpler check, but it looks like the current semantics is fairly common.

rossberg:
Regarding C implementations of memmove: C does not specify any order, instead it says the semantics is as if copying to a temporary buffer first. More importantly, both data races and OOB reads/writes are undefined behaviour in C. So AFAICT, the only way to tell the difference between the two conditions is through UB, which means it should be immaterial.

binji:
True, but the fact that C library implementers have a choice may mean they are choosing an ordering based on what they found to be faster.

PR #91, opened 2019-05-24
[spec] Runtime format of element and data segments

binji:
This is copied and modified from this closed PR:

https://github.com/WebAssembly/bulk-memory-operations/pull/19


binji:
Yes, that makes sense. I'll update the PR

PR #92, opened 2019-05-31
[spec] Add memory.init and data.drop for exec

binji:


binji:
I'll land this now, if there's any feedback I'll address in a later PR.

document/core/exec/instructions.rst
```diff
@@ -689,6 +689,130 @@ Memory Instructions
...
+4. If :math:`a^? = \epsilon`, then:
+
+   a. Trap.
+
+5. Replace :math:`F.\AMODULE.\MIDATAS[x]` with :math:`\epsilon`.
+
+.. math::
+   \begin{array}{lcl@{\qquad}l}
+   F; (\DATADROP~x) &\stepto& F'; \epsilon
+     & (\iff F' = F \with \AMODULE.\MIDATAS[x] = \epsilon) \\
```
rossberg: Ah, unfortunately, this isn't enough. The same module may occur in other frames or function instances, and this just updates the one "copy", although we'll need to update all of them.

The underlying problem is that modules are stateful now, which they weren't before. Since they are global, that means that they will have to become part of the store. That is, we need to reference them via an indirection through a *module address* that indexes into a new store field for allocated modules, and updates to the module are done as updates to the store, like other writes. That's a bit of an annoying change, but unavoidable.

binji: Ah, right.

binji: Actually, I think it may be simpler than this. If the data and element segments are already in the store, they should be optional instead of their addresses in the frame.

binji: See #93.

Issue #15, opened 2018-05-10
Import/export passive segments?


binji:
@rossberg mentioned this [here](https://github.com/WebAssembly/bulk-memory-operations/pull/12#discussion_r187429452). Have we discussed importing and exporting passive data and element segments?

Perhaps this would be useful for dynamic linking, though I'm not quite sure how.

rossberg:
One use case could e.g. be the ability to create a module that provides a specific set of string constants as data segments. Then those would be reusable without also having to share a memory or even a memory management policy.

lukewagner:
It sounds possible, but it's not quite a trivial addition (e.g., a new JS API object, probably needing accessors for the contents of the segment), so it'd be nice to have more of a clear need for it.  Also, once we have multi-memory/table, one could achieve the "exported constants without memory management policy" use case by exporting memories/tables (initialized by active non-exported data/elem segments), or even just globals (allowing each one to be named).

PR #93, opened 2019-06-08
[spec] Use optional data instance, not data address

binji:
data.drop must update the store, not the frame. There may be multiple
copies of the frame, so any updates will only update once. We can make
sure that all copies are updated by using an indirection through data
addresses and updating the store instead.

See discussion in PR #92.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -704,29 +704,31 @@ Memory Instructions
 
 5. Let :math:`\X{mem}` be the :ref:`memory instance <syntax-meminst>` :math:`S.\SMEMS[\X{ma}]`.
 
-6. Assert: due to :ref:`validation <valid-memory.init>`, :math:`F.\AMODULE.\MIDATAS[x]` exists.
+6. Assert: due to :ref:`validation <valid-data.init>`, :math:`F.\AMODULE.\MIDATAS[x]` exists.
 
-7. Let :math:`\X{da}^?` be the optional :ref:`data address <syntax-dataaddr>` :math:`F.\AMODULE.\MIDATAS[x]`.
+7. Let :math:`\X{da}` be the :ref:`data address <syntax-dataaddr>` :math:`F.\AMODULE.\MIDATAS[x]`.
 
-8. If :math:`\X{da}^? = \epsilon`, then:
+8. Assert: due to :ref:`validation <valid-data.init>`, :math:`S.\SDATA[\X{da}]` exists.
+
+9. Let :math:`\X{data}^?` be the optional :ref:`data instance <syntax-datainst>` :math:`S.\SDATA[\X{da}]`.
+
+10. If :math:`\X{data}^? = \epsilon`, then:
 
    a. Trap.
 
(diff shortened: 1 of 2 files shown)


document/core/exec/instructions.rst
```diff
@@ -765,18 +768,19 @@ Memory Instructions
    \end{array} \\
    \\ \qquad
      \begin{array}[t]{@{}r@{~}l@{}}
-     (\iff & F.\AMODULE.\MIDATAS[x] \ne \epsilon \\
-     \wedge & b = \SDATA[F.\AMODULE.\MIDATAS[x]].\DIINIT[s]) \\
+     (\iff & S.\SDATA[F.\AMODULE.\MIDATAS[x]] \ne \epsilon \\
```
rossberg: You can actually drop this extra condition, because it is implied by the others, which can only hold if it is not epsilon.

document/core/exec/instructions.rst
```diff
@@ -765,18 +768,19 @@ Memory Instructions
...
    \begin{array}{lcl@{\qquad}l}
    S; F; (\I32.\CONST~d)~(\I32.\CONST~(s)~(\I32.\CONST~0)~(\MEMORYINIT~x) &\stepto& S; F; \epsilon
    \end{array}
    \\ \qquad
      \begin{array}[t]{@{}r@{~}l@{}}
-     (\iff & F.\AMODULE.\MIDATAS[x] \ne \epsilon \\
-     \wedge & d \leq |\SMEMS[F.\AMODULE.\MIMEMS[0]]| \\
-     \wedge & s \leq |\SDATA[F.\AMODULE.\MIDATAS[x]]|) \\
+     (\iff & d \leq |S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA| \\
+     \wedge & S.\SDATA[F.\AMODULE.\MIDATAS[x]] \ne \epsilon \\
```
rossberg: Same here.

document/core/exec/runtime.rst
```diff
@@ -72,8 +72,8 @@ Syntactically, the store is defined as a :ref:`record <notation-record>` listing
      \STABLES & \tableinst^\ast, \\
      \SMEMS & \meminst^\ast, \\
      \SGLOBALS & \globalinst^\ast, \\
-     \SELEM & \eleminst^\ast, \\
-     \SDATA & \datainst^\ast ~\} \\
+     \SELEM & (\eleminst^?)^\ast, \\
+     \SDATA & (\datainst^?)^\ast ~\} \\
      \end{array}
    \end{array}
```
rossberg: Maybe add an explanatory sentence here like the following:
```
Element and data segments can be dropped by the owning module, in which case the respective instance is replaced with :math:`\epsilon`.
It is an invariant of the semantics that no element or data instance is :ref:`addressed <syntax-addr>` from anywhere else but the owning module instances.
```

binji:
Right, that seems to be a safe assumption for now (as discussed in #15). I started working on moving the module to the store, but it's a much larger change -- I'd rather do that when/if we need to.

PR #95, opened 2019-06-11
[spec] Add `table.init`, `elem.drop` exec text

binji:
This also adds `table.set` and `table.get` administrative instructions.
Only `table.set` is used for now, but `table.copy` will use both.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -830,6 +830,245 @@ Memory Instructions
    \end{array}
 
 
+.. index:: table instruction, table index, store, frame, address, table address, table instance, function element, element address, element instance, value type
+   pair: execution; instruction
+   single: abstract syntax; instruction
+.. _exec-instr-table:
+
+Table Instructions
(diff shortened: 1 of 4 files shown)


document/core/exec/instructions.rst
```diff
@@ -830,6 +830,245 @@ Memory Instructions
...
+
+    h. Execute the instruction :math:`\TABLEINIT~x`.
+
+.. math::
+   ~\\[-1ex]
+   \begin{array}{l}
+   \begin{array}{lcl@{\qquad}l}
+   S; F; (\I32.\CONST~d)~(\I32.\CONST~s)~(\I32.\CONST~(n+1))~(\TABLEINIT~x) &\stepto& S; F;
+     \begin{array}[t]{@{}l@{}}
+     (\I32.\CONST~d)~\funcelem~\TABLESET \\
```
rossberg: I'm afraid this requires that you make `\funcelem` an administrative instruction as well.

binji:
OK, I'll land this and fix the `\funcelem` issue in the next PR.

PR #101, opened 2019-06-28
[spec] Update text for zero-byte OOB case

binji:
Also fix some typos from previous CL (see macros.def)

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -645,13 +645,7 @@ Memory Instructions
 
 11. Pop the value :math:`\I32.\CONST~i` from the stack.
 
-12. If :math:`n` is :math:`0`, then:
-
-    a. If :math:`i` is larger than the length of :math:`\X{mem}.\MIDATA`, then:
-
-       i. Trap.
-
-13. Else:
+12. Else:
 
     a. Push the value :math:`\I32.\CONST~i` to the stack.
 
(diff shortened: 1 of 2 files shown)


document/core/exec/instructions.rst
```diff
@@ -773,13 +750,8 @@ Memory Instructions
...
+   S; F; (\I32.\CONST~d)~(\I32.\CONST~s)~(\I32.\CONST~0)~(\MEMORYINIT~x) &\stepto& S; F; \epsilon
    \end{array}
-   \\ \qquad
-     \begin{array}[t]{@{}r@{~}l@{}}
-     (\iff & d \leq |S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA| \\
-     \wedge & s \leq |S.\SDATA[F.\AMODULE.\MIDATAS[x]].\DIINIT|) \\
-     \end{array}
    \\[1ex]
    \begin{array}{lcl@{\qquad}l}
    S; F; (\I32.\CONST~d)~(\I32.\CONST~s)~(\I32.\CONST~n)~(\MEMORYINIT~x) &\stepto& S; F; \TRAP
```
rossberg: I believe this case needs to be removed now.

document/core/exec/instructions.rst
```diff
@@ -922,11 +884,6 @@ Table Instructions
...
    S; F; (\I32.\CONST~d)~(\I32.\CONST~(s)~(\I32.\CONST~0)~(\TABLEINIT~x) &\stepto& S; F; \epsilon
    \end{array}
-   \\ \qquad
-     \begin{array}[t]{@{}r@{~}l@{}}
-     (\iff & d \leq |S.\STABLES[F.\AMODULE.\MITABLES[0]].\TIELEM| \\
-     \wedge & s \leq |S.\SELEM[F.\AMODULE.\MIELEMS[x]].\EIINIT|) \\
-     \end{array}
    \\[1ex]
    \begin{array}{lcl@{\qquad}l}
    S; F; (\I32.\CONST~d)~(\I32.\CONST~s)~(\I32.\CONST~n)~(\TABLEINIT~x) &\stepto& S; F; \TRAP
```
rossberg: Same here.

document/core/exec/instructions.rst
binji: I don't think it can, since we need to trap if the first condition is not met, right? (i.e. the source offset is out of bounds).

rossberg: Ah, you're right, of course. You might want to reorder the rules, since the "otherwise" is a bit weird when the preceding rule has no side condition. Maybe put the zero rule first? (Same below)

binji: Done.

PR #102, opened 2019-06-28
[interp] Don't trap on zero-bytes out-of-bounds

binji:


binji:
Gonna land now, will follow up with comments in a future PR.

PR #104, opened 2019-07-11
Polish table.init exec spec

gahaas:
This PR polishes the execution spec of `table.init`:
* Use `tab` as the name of the table, to be consistent with other instructions like `table.set`.
* Use more describing names for stack values: `dst`, `src`, and `cnt`, instead of `d`, `s`, `n`, respectively.
* Fix step numbering.
* Add return when `cnt = 0`.
* Add trap when `src` is out of bounds.
* Fix the call to `table.set`, i.e. push the `funcelem` on the stack, and call it with the correct immediate.
* In the formalism part:
  * Change `cnt` to `(cnt+1)` in the trap case.
  * Change the successor state in the not-trap cases from `S` to `S'`, as it is done for other instructions.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -863,7 +863,7 @@ Table Instructions
 
 21. Push the value :math:`\funcelem` to the stack.
 
-22. Execute the instruction :math:`\TABLESET~x`.
+22. Execute the instruction :math:`\TABLESET`.
 
 23. Push the value :math:`\I32.\CONST~(dst+1)` to the stack.
 
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -877,7 +877,7 @@ Table Instructions
    ~\\[-1ex]
    \begin{array}{l}
    \begin{array}{lcl@{\qquad}l}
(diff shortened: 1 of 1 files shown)


document/core/exec/instructions.rst
```diff
@@ -835,58 +835,66 @@ Table Instructions
...
+25. Push the value :math:`\I32.\CONST~(cnt-1)` to the stack.
+
+26. Execute the instruction :math:`\TABLEINIT~x`.
 
 .. math::
    ~\\[-1ex]
    \begin{array}{l}
    \begin{array}{lcl@{\qquad}l}
-   S; F; (\I32.\CONST~d)~(\I32.\CONST~(s)~(\I32.\CONST~0)~(\TABLEINIT~x) &\stepto& S; F; \epsilon
+   S; F; (\I32.\CONST~dst)~(\I32.\CONST~(src)~(\I32.\CONST~0)~(\TABLEINIT~x) &\stepto& S'; F; \epsilon
```
rossberg:
```suggestion
   S; F; (\I32.\CONST~dst)~(\I32.\CONST~src)~(\I32.\CONST~0)~(\TABLEINIT~x) &\stepto& S'; F; \epsilon
```

PR #106, opened 2019-07-18
[spec] Add `memory.copy` exec text

binji:
Also clean up some `table.copy` text.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -649,21 +649,23 @@ Memory Instructions
 
     a. Return.
 
-13. Else:
+13. Push the value :math:`\I32.\CONST~i` to the stack.
 
-    a. Push the value :math:`\I32.\CONST~i` to the stack.
+14. Push the value :math:`\val` to the stack.
 
-    b. Push the value :math:`\val` to the stack.
+15. Execute the instruction :math:`\I32\K{.}\STORE\K{8}~\{ \OFFSET~0, \ALIGN~0 \}`.
 
-    c. Execute the instruction :math:`\I32\K{.}\STORE\K{8}~\{ \OFFSET~0, \ALIGN~0 \}`.
+16. If :math:`n = 1`, then:
 
-    d. Push the value :math:`\vconst_\I32(i+1)` to the stack.
+    a. Return.
 
-    e. Push the value :math:`\val` to the stack.
+17. Push the value :math:`\vconst_\I32(i+1)` to the stack.
 
-    f. Push the value :math:`\I32.\CONST~(n-1)` to the stack.
+18. Push the value :math:`\val` to the stack.
 
-    g. Execute the instruction :math:`\MEMORYFILL`.
+19. Push the value :math:`\I32.\CONST~(n-1)` to the stack.
+
+20. Execute the instruction :math:`\MEMORYFILL`.
 
 .. math::
    \begin{array}{l}
@@ -671,12 +673,18 @@ Memory Instructions
    S; F; (\I32.\CONST~i)~\val~(\I32.\CONST~0)~(\MEMORYFILL) &\stepto& S; F; \epsilon
    \end{array} \\
    \begin{array}{lcl@{\qquad}l}
-   S; F; (\I32.\CONST~i)~\val~(\I32.\CONST~(n+1))~(\MEMORYFILL) &\stepto& S; F;
-     \begin{array}[t]{@{}l@{}}
+   S; F; (\I32.\CONST~i)~\val~(\I32.\CONST~1)~(\MEMORYFILL) &\stepto& S; F;
      (\I32.\CONST~i)~\val~(\I32\K{.}\STORE\K{8}~\{ \OFFSET~0, \ALIGN~0 \}) \\
-     (\vconst_\I32(i+1))~\val~(\I32.\CONST~n)~(\MEMORYFILL) \\
-     \end{array} \\
    \end{array} \\
+   \begin{array}{lcl@{\qquad}l}
+   S; F; (\I32.\CONST~i)~\val~(\I32.\CONST~n)~(\MEMORYFILL) &\stepto& S; F;
+     \begin{array}[t]{@{}l@{}}
+     (\I32.\CONST~i)~\val~(\I32.\CONST~1)~(\MEMORYFILL) \\
+     (\vconst_\I32(i+1))~\val~(\I32.\CONST~(n-1))~(\MEMORYFILL) \\
+     \end{array} \\
+   \end{array}
+   \\ \qquad
+     (\iff n > 1) \\
    \end{array}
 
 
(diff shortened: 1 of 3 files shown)


binji:
See https://webassembly.github.io/bulk-memory-operations/core/exec/instructions.html.

document/core/exec/instructions.rst
```diff
@@ -801,6 +801,93 @@ Memory Instructions
...
+     \begin{array}[t]{@{}l@{}}
+     (\I32.\CONST~dst) \\
+     (\I32.\CONST~src)~(\I32\K{.}\LOAD\K{8\_u}~\{ \OFFSET~0, \ALIGN~0 \}) \\
+     (\I32\K{.}\STORE\K{8}~\{ \OFFSET~0, \ALIGN~0 \}) \\
+     (\I32.\CONST~(dst+1))~(\I32.\CONST~(src+1))~(\I32.\CONST~cnt)~\MEMORYCOPY \\
+     \end{array} \\
+   \end{array} \\
+   \\ \qquad
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\iff & dst < src)
```
rossberg: Did you mean `src < dst`? (Also, I'd probably make that <=)

document/core/exec/instructions.rst
```diff
@@ -801,6 +801,93 @@ Memory Instructions
...
+   S; F; (\I32.\CONST~dst)~(\I32.\CONST~src)~(\I32.\CONST~0)~\MEMORYCOPY &\stepto& S; F; \epsilon
+   \end{array}
+   \\[1ex]
+   \begin{array}{lcl@{\qquad}l}
+   S; F; (\I32.\CONST~dst)~(\I32.\CONST~src)~(\I32.\CONST~(cnt+1))~\MEMORYCOPY &\stepto& S; F;
+     \begin{array}[t]{@{}l@{}}
+     (\I32.\CONST~dst) \\
+     (\I32.\CONST~src)~(\I32\K{.}\LOAD\K{8\_u}~\{ \OFFSET~0, \ALIGN~0 \}) \\
+     (\I32\K{.}\STORE\K{8}~\{ \OFFSET~0, \ALIGN~0 \}) \\
+     (\I32.\CONST~(dst+1))~(\I32.\CONST~(src+1))~(\I32.\CONST~cnt)~\MEMORYCOPY \\
```
rossberg: Crap, one annoyance I just realised is that dst+1 (and src+1) may overflow the i32 range, in which case `(i32.const (dst+1))` is not well-formed syntactically. That problem also applies to all other bulk operators. I think that requires a modification, e.g.:
```
S; F; (i32.const dst) (i32.const src) (i32.const cnt) memory.copy ~>
S; F; (i32.const dst) (i32.const src) (i32.load8_u {..}) (i32.store8 {...})
      const_i32(dst+1) const_i32(dst+1) (i32.const cnt) memory.copy
```
with the auxiliary meta function
```
const_t(x) = (t.const x)  (iff x is well-formed for t)
const_t(x) = trap  (otherwise)
```
Not sure where to best put that function, since it will be needed for the other bulk ops as well. Perhaps it could go into a "Conventions" subsection for the Values section in runtime.rst.

document/core/exec/instructions.rst
```diff
@@ -801,6 +801,93 @@ Memory Instructions
...
+   \\ \qquad
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\iff & dst < src)
+     \end{array}
+   \\[1ex]
+   \begin{array}{lcl@{\qquad}l}
+   S; F; (\I32.\CONST~dst)~(\I32.\CONST~src)~(\I32.\CONST~(cnt+1))~\MEMORYCOPY &\stepto& S; F;
+     \begin{array}[t]{@{}l@{}}
+     (\I32.\CONST~(dst+cnt-1)) \\
+     (\I32.\CONST~(src+cnt-1))~(\I32\K{.}\LOAD\K{8\_u}~\{ \OFFSET~0, \ALIGN~0 \}) \\
```
rossberg: Same here, these computations might overflow.

rossberg: Also, I think the -1 is off by one, since the rule is for cnt+1.

document/core/exec/instructions.rst
rossberg: Actually, the trap trick would be wrong when cnt happens to be 0... Hm, I can't think of a way to avoid an extra rule.

lars-t-hansen: FWIW, SpiderMonkey goes out of its way to avoid trapping just because a computation overflows, dropping into 64-bit arithmetic in some cases.  This goes for all the bulk ops.  Mostly this is to be able to trap midway through an operation though, without doing bounds checking (and checking for overflow) for every element.

document/core/exec/instructions.rst
binji: I think this is correct; we copy backward only when `src < dst` (but you're right this case should be `dst <= src`)

document/core/exec/instructions.rst
binji: Ah, good point. I'll add it.

binji:
Updated published version: https://webassembly.github.io/bulk-memory-operations/core/exec/instructions.html#memory-instructions

document/core/exec/instructions.rst
binji: @rossberg not sure what you mean, what's wrong with `cnt` = 0?

document/core/exec/instructions.rst
binji: Done.

document/core/exec/instructions.rst
rossberg: The problem for cnt=0 is that this formulation now unconditionally traps after processing address 2^32-1, even when the remaining cnt is zero after that, which should be okay. So e.g. `(memory.copy (i32.const 2^32-1) (i32.const 0) (i32.const 1))` would incorrectly trap after writing the byte.

Unfortunately, I can't think of a way of handling that correctly without multiple extra rules, essentially duplicating the logic of the existing cases. The only other alternative is introducing an administrative version of each bulk instruction that incorporates the running indices as plain natural numbers and thus side-steps the value range limitation. Both solutions seem really annoying.

It's also unfortunate that it's practically impossible to add tests for these edge cases, even though engines will likely run into similar issues, as @lars-t-hansen pointed out.

binji: I see what you mean. I've added a `cnt=1` case for this, which I think works OK without being too verbose. PTAL

binji:
Updated published version again (same link)

document/core/exec/instructions.rst
```diff
@@ -749,20 +763,27 @@ Memory Instructions
...
+     (\I32.\CONST~d)~(\I32.\CONST~s)~(\I32.\CONST~1)~(\MEMORYINIT~x) \\
+     (\vconst_\I32(d+1))~(\vconst_\I32(s+1))~(\I32.\CONST~(n-1))~(\MEMORYINIT~x) \\
      \end{array} \\
-   \end{array} \\
+   \end{array}
+   \\ \qquad
+     (\iff n > 1)
+   \\[1ex]
+   \begin{array}{lcl@{\qquad}l}
+   S; F; (\I32.\CONST~d)~(\I32.\CONST~s)~(\I32.\CONST~1)~(\MEMORYINIT~x) &\stepto& S; F;
```
rossberg: Perhaps put this case before the `n>1` one.

document/core/exec/instructions.rst
```diff
@@ -1045,36 +1088,47 @@ Table Instructions
...
+     (\I32.\CONST~dst)~(\I32.\CONST~src)~(\I32.\CONST~1)~(\TABLEINIT~x) \\
+     (\vconst_\I32(dst+1))~(\vconst_\I32(src+1))~(\I32.\CONST~(cnt-1))~(\TABLEINIT~x) \\
      \end{array} \\
-   \end{array} \\
+   \end{array}
+   \\ \qquad
+     (\iff cnt > 1)
+   \\[1ex]
+   \begin{array}{lcl@{\qquad}l}
+   S; F; (\I32.\CONST~dst)~(\I32.\CONST~src)~(\I32.\CONST~1)~(\TABLEINIT~x) &\stepto& S; F;
```
rossberg: Move before `n>1` case.

binji:
Updated prose to match the formal pretty closely, though it's still not an exact match.

Issue #98, opened 2019-06-20
Change passive element segment encoding


binji:
We discussed this at the 2019 June CG meeting, and decided to update the passive element segment encoding.

See https://github.com/WebAssembly/meetings/blob/master/2019/CG-06.md#bulk-memory--hr (particularly the slides).

binji:
Proposed encoding (from slide 14):

| Flags Value | Segment Type | Externval Index / Element Expression | Table 0 / Table index field |
| - | - | - | - |
| 0 | Active | Externval index | Table 0 |
| 1 | Active | Element expression | Table 0 |
| 2 | Active | Externval index | Table index field |
| 3 | Active | Element expression | Table index field |
| 4 | Passive | Externval index | N/A |
| 5 | Passive | Element expression | N/A |
| 6 | Declared | Externval index | N/A |

Memory layout (from slide 16)

| Name | flags | - | - | - | - | - |
| - | - | - | - | - | - | - |
| Active, externval, table 0 | 0 | offset | func 0 | func 1 | func 2 | func 3 |
| Active, elemexpr, table 0 | 1 | offset | ref.func | func 0 | end | |
| Active, externval, table index | 2 | offset | table index | externkind | index 0 | index 1 |
| Active, elemexpr, table index | 3 | offset | table index | ref.func | func 0 | end |
| Passive, externval | 4 | elemtype | externkind | index 0 | index 1 | index 2 |
| Passive, elemexpr | 5 | elemtype | ref.func | func 0 | end | |
| Declared, externval | 6 | externkind | index 0 | index 1 | index 2 | index 3 |

Any concerns with this?

lars-t-hansen:
No concerns.

titzer:
I guess this is a bikeshedding thread. :-)

What about making the low-order bit of the flags field for the first table indicate whether a table index immediate follows? That way, no table lookup is necessary for a decoder to skip to the next entry, just a bit check.

rossberg:
I generally would prefer compatibility with the flag layout for data segments, i.e., use more bits, but have the same bits mean the same thing. I also wouldn't mind a bit more separation of te various dimensions.

For data segments we have:

| Bit | Meaning |
| - | - |
| 0 | is passive |
| 1 | is zero idx (if bit 0 clear) or clear otherwise |

For elements that would suggest:

| Bit | Meaning |
| - | - |
| 0 | is passive |
[...]

PR #108, opened 2019-07-31
Change element segment encoding

gahaas:
This PR implements the changes to the element segment encoding discussed in https://github.com/WebAssembly/bulk-memory-operations/issues/98.

I don't mention declared segments yet, because they don't exist in this proposal. I also don't mention
the externkind but immediately set it to 0x00, which is the value for external functions anyways.

### DIFF
--- a/document/core/binary/modules.rst
+++ b/document/core/binary/modules.rst
@@ -331,14 +331,14 @@ It decodes into a vector of :ref:`element segments <syntax-elem>` that represent
        &\Rightarrow& \{ \ETABLE~0, \EOFFSET~o, \ETYPE~\FUNCREF, \EINIT~((\REFFUNC~y)~\END)^\ast \} \\ &&|&
      \hex{01}~~\hex{00}~~y^\ast{:}\Bvec(\Bfuncidx)
        &\Rightarrow& \{ \ETYPE~\FUNCREF, \EINIT~((\REFFUNC~y)~\END)^\ast \} \\ &&|&
-     \hex{02}~~t{:}\Btableidx~~o{:}\Bexpr~~\hex{00}~~y^\ast{:}\Bvec(\Bfuncidx)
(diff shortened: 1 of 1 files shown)


document/core/binary/modules.rst
```diff
@@ -327,12 +327,18 @@ It decodes into a vector of :ref:`element segments <syntax-elem>` that represent
...
      \X{seg}^\ast{:}\Bsection_9(\Bvec(\Belem)) &\Rightarrow& \X{seg} \\
    \production{element segment} & \Belem &::=&
-     \hex{00}~~e{:}\Bexpr~~y^\ast{:}\Bvec(\Bfuncidx)
-       &\Rightarrow& \{ \ETABLE~0, \EOFFSET~e, \EINIT~((\REFFUNC~y)~\END)^\ast \} \\ &&|&
-     \hex{01}~~\X{et}:\Belemtype~e^\ast{:}\Bvec(\Belemexpr)
+     \hex{00}~~o{:}\Bexpr~~y^\ast{:}\Bvec(\Bfuncidx)
+       &\Rightarrow& \{ \ETABLE~0, \EOFFSET~o, \ETYPE~\FUNCREF, \EINIT~((\REFFUNC~y)~\END)^\ast \} \\ &&|&
+     \hex{01}~~\hex{00}~~y^\ast{:}\Bvec(\Bfuncidx)
+       &\Rightarrow& \{ \ETYPE~\FUNCREF, \EINIT~((\REFFUNC~y)~\END)^\ast \} \\ &&|&
+     \hex{02}~~t{:}\Btableidx~~o{:}\Bexpr~~\hex{00}~~y^\ast{:}\Bvec(\Bfuncidx)
```
rossberg: Convention is to use `x` for indices (`t` is for types). Same below.

gahaas: Done

PR #109, opened 2019-08-05
New element section encoding in the interpreter

gahaas:
This PR implements https://github.com/WebAssembly/bulk-memory-operations/issues/98:
* Extend the interpreter to handle the new element segment encoding;
* Add tests for the new element segment encoding;
* Adjust the text format specification;

The text format for case `0x1` needs the `func` keyword as a separator between a potential segment identifier and the function indices.

The text format needs a more verbose table index to distinguish between a potential segment identifier and the table index in case `0x2`. Alternatively we could just disallow table identifiers for active segments, since they are not useful anyways.

also-by: thibaudm@chromium.org

### DIFF
--- a/document/core/syntax/modules.rst
+++ b/document/core/syntax/modules.rst
@@ -261,7 +261,7 @@ The |MELEM| component of a module defines a vector of element segments. Each act
 .. math::
    \begin{array}{llll}
    \production{element segment} & \elem &::=&
-     \{ \ETABLE~\tableidx, \EOFFSET~\expr, \EINIT~\vec(\elemexpr) \} \\&&|&
+     \{ \ETABLE~\tableidx, \EOFFSET~\expr, \ETYPE~\elemtype, \EINIT~\vec(\elemexpr) \} \\&&|&
      \{ \ETYPE~\elemtype, \EINIT~\vec(\elemexpr) \} \\
    \production{elemexpr} & \elemexpr &::=&
      \REFNULL~\END \\&&|&
--- a/document/core/text/modules.rst
+++ b/document/core/text/modules.rst
@@ -484,15 +484,15 @@ Element segments allow for an optional :ref:`table index <text-tableidx>` to ide
    \begin{array}{llclll}
    \production{element segment} & \Telem_I &::=&
      \text{(}~\text{elem}~~\Tid^?~~e{:}\Toffset_I~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{ \ETABLE~0, \EOFFSET~e, \EINIT~y^\ast \} \\ &&|&
+       \Rightarrow\quad \{ \ETABLE~0, \EOFFSET~e, \ETYPE~\FUNCREF, \EINIT~y^\ast \} \\ &&|&
      \text{(}~\text{elem}~~\Tid^?~~\text{(}~\text{table}~~x{:}\Ttableidx_I ~\text{)}~~e{:}\Toffset_I~~\text{func}~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{ \ETABLE~x, \EOFFSET~e, \EINIT~y^\ast \} \\ &&|&
+       \Rightarrow\quad \{ \ETABLE~x, \EOFFSET~e, \ETYPE~\FUNCREF, \EINIT~y^\ast \} \\ &&|&
      \text{(}~\text{elem}~~\Tid^?~~e{:}\Toffset_I~~y^\ast{:}\Tvec(\Texpr_I)~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{ \ETABLE~0, \EOFFSET~e, \EINIT~y^\ast \} \\ &&|&
+       \Rightarrow\quad \{ \ETABLE~0, \EOFFSET~e, \ETYPE~\FUNCREF, \EINIT~y^\ast \} \\ &&|&
      \text{(}~\text{elem}~~\Tid^?~~\text{(}~\text{table}~~x{:}\Ttableidx_I ~\text{)}~~e{:}\Toffset_I~~et{:}\Telemtype~~y^\ast{:}\Tvec(\Texpr_I)~\text{)} \\ &&& \qquad
        \Rightarrow\quad \{ \ETABLE~x, \EOFFSET~e, \ETYPE~et, \EINIT~y^\ast \} \\ &&|&
      \text{(}~\text{elem}~~\Tid^?~~\text{func}~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{\EINIT~y^\ast \} \\ &&|&
+       \Rightarrow\quad \{\ETYPE~\FUNCREF, \EINIT~y^\ast \} \\ &&|&
      \text{(}~\text{elem}~~\Tid^?~~et{:}\Telemtype~~y^\ast{:}\Tvec(\Texpr_I)~\text{)} \\ &&& \qquad
        \Rightarrow\quad \{\ETYPE~et,\EINIT~y^\ast \} \\
    \production{offset}  & \Toffset_I &::=&
--- a/document/core/text/modules.rst
+++ b/document/core/text/modules.rst
@@ -493,20 +493,13 @@ Element segments allow for an optional :ref:`table index <text-tableidx>` to ide
 .. math::
    \begin{array}{llclll}
    \production{element segment} & \Telem_I &::=&
-     \text{(}~\text{elem}~~\Tid^?~~e{:}\Toffset_I~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{ \ETABLE~0, \EOFFSET~e, \ETYPE~\FUNCREF, \EINIT~y^\ast \} \\ &&|&
-     \text{(}~\text{elem}~~\Tid^?~~e{:}\Toffset_I~~\{et, init\}{:}\Tfunclist_I~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{ \ETABLE~0, \EOFFSET~e, \ETYPE~et, \EINIT~init \} \\ &&|&
-     \text{(}~\text{elem}~~\Tid^?~~\text{(}~\text{table}~~x{:}\Ttableidx_I ~\text{)}~~e{:}\Toffset_I~~\{et, init\}{:}\Tfunclist_I~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{ \ETABLE~x, \EOFFSET~e, \ETYPE~et, \EINIT~init \} \\ &&|&
-     \text{(}~\text{elem}~~\Tid^?~~\{et, init\}{:}\Tfunclist_I~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{\ETYPE~et,\EINIT~init \} \\
-   \production{offset}  & \Toffset_I &::=&
-     \text{(}~\text{offset}~~e{:}\Texpr_I~\text{)} \qquad\Rightarrow\quad \EOFFSET~e \\ &&|&
-     e{:}\Texpr_I \qquad\Rightarrow\quad \EOFFSET~e \\
-   \production{funclist}  & \Tfunclist_I &::=&
-     \text{func}~~y^\ast{:}\Tvec(\Tfuncidx_I) \qquad\Rightarrow\quad \{ \ETYPE~\FUNCREF, \EINIT~y^\ast \} \\ &&|&
-     et{:}\Telemtype~~y^\ast{:}\Tvec(\Texpr_I) \qquad\Rightarrow\quad \{ \ETYPE~et, \EINIT~y^\ast \} \\
+     \text{(}~\text{elem}~~\Tid^?~~\text{(}~\text{table}~~x{:}\Ttableidx_I ~\text{)}~~\text{(}~\text{offset}~~e{:}\Texpr_I~\text{)}~~(et, y^\ast){:}\Telemlist~\text{)} \\ &&& \qquad
+       \Rightarrow\quad \{ \ETABLE~x, \EOFFSET~e, \ETYPE~et, \EINIT~y^\ast \} \\ &&|&
+     \text{(}~\text{elem}~~\Tid^?~~(et, y^\ast){:}\Telemlist~\text{)} \\ &&& \qquad
+       \Rightarrow\quad \{\ETYPE~et,\EINIT~y^\ast \} \\
+   \production{elemlist}  & \Telemlist &::=&
+     \text{func}~~y^\ast{:}\Tvec(\Tfuncidx_I) \qquad\Rightarrow\quad ( \ETYPE~\FUNCREF, \EINIT~y^\ast ) \\ &&|&
+     et{:}\Telemtype~~y^\ast{:}\Tvec(\Texpr_I) \qquad\Rightarrow\quad ( \ETYPE~et, \EINIT~y^\ast ) \\
    \end{array}
 
 .. note::
@@ -526,14 +519,14 @@ As an abbreviation, a single instruction may occur in place of the offset:
      \text{(}~\text{offset}~~\Tinstr~\text{)}
    \end{array}
 
-Also, the table index can be omitted, defaulting to :math:`\T{0}`.
+Also, the table index can be omitted, defaulting to :math:`\T{0}`. If the table index is omitted, also the :math:`\text{func}` keyword can be omitted.
 
 .. math::
    \begin{array}{llclll}
    \production{element segment} &
     \text{(}~\text{elem}~~\text{(}~\text{offset}~~\Texpr_I~\text{)}~~\dots~\text{)}
        &\equiv&
-     \text{(}~\text{elem}~~0~~\text{(}~\text{offset}~~\Texpr_I~\text{)}~~\dots~\text{)}
+     \text{(}~\text{elem}~~\text{(}~\text{table}~~\text{0}~\text{)}~~\text{(}~\text{offset}~~\Texpr_I~\text{)}~~\dots~\text{)}
    \end{array}
 
 As another abbreviation, element segments may also be specified inline with :ref:`table <text-table>` definitions; see the respective section.
--- a/document/core/text/modules.rst
+++ b/document/core/text/modules.rst
@@ -293,6 +293,16 @@ An :ref:`element segment <text-elem>` can be given inline with a table definitio
        (\iff \Tid' = \Tid^? \neq \epsilon \vee \Tid' \idfresh) \\
    \end{array}
 
+.. math::
+   \begin{array}{llclll}
+   \production{module field} &
+     \text{(}~\text{table}~~\Tid^?~~\Telemtype~~\text{(}~\text{elem}~~x^n{:}\Tvec(\Texpr)~\text{)}~~\text{)} \quad\equiv \\ & \qquad
+       \text{(}~\text{table}~~\Tid'~~n~~n~~\Telemtype~\text{)}~~
+       \text{(}~\text{elem}~~\Tid'~~\text{(}~\text{i32.const}~~\text{0}~\text{)}~~\Tvec(\Texpr)~\text{)}
+       \\ & \qquad\qquad
+       (\iff \Tid' = \Tid^? \neq \epsilon \vee \Tid' \idfresh) \\
+   \end{array}
+
 Tables can be defined as :ref:`imports <text-import>` or :ref:`exports <text-export>` inline:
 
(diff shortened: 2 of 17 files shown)


interpreter/syntax/ast.ml
```diff
@@ -140,18 +140,23 @@ and memory' =
...
 
-type table_segment = (elem list, elem_type) segment
-type memory_segment = (string, unit) segment
+
+type table_segment = table_segment' Source.phrase
+and table_segment' =
+  | ActiveWithRefs of {index : var; offset : const; etype : elem_type; init : elem list}
+  | ActiveWithIndices of {index : var; offset : const; init : elem list}
+  | PassiveWithRefs of {etype : elem_type; data : elem list}
+  | PassiveWithIndices of {data : elem list}
```
rossberg: AFAICS, there is no value in carrying this distinction into the AST, it just complicates everything downstream. Instead, simply decode the index cases as (a short-hand for) ref cases, converting the extern_kind into a suitable reftype. In the encoder, pick index form when possible.

If I'm not mistaken, that would also make the duplication of the segment types and code unnecessary, but I may be missing something.

interpreter/syntax/types.ml
```diff
@@ -2,6 +2,7 @@
 
 type value_type = I32Type | I64Type | F32Type | F64Type
 type elem_type = FuncRefType
+type extern_kind = FuncKind
```
rossberg: With the suggestion above, this wouldn't be needed.

interpreter/text/parser.mly
```diff
@@ -221,6 +221,9 @@ value_type_list :
 elem_type :
   | FUNCREF { FuncRefType }
 
+extern_kind :
```
rossberg: Nit: move to section on tables etc. Also, rename to `elem_kind`.

interpreter/text/parser.mly
```diff
@@ -574,34 +580,44 @@ elemref :
...
       fun c -> ignore ($3 c anon_elem bind_elem);
-      fun () -> Passive {etype = $4; data = $5 c} @@ at }
-  | LPAR ELEM bind_var var offset active_elemref_list RPAR
-    { let at = at () in
-      fun c -> ignore (bind_elem c $3);
       fun () ->
-      Active {index = $4 c table; offset = $5 c; init = $6 c func} @@ at }
-  | LPAR ELEM var offset active_elemref_list RPAR
+        ActiveWithIndices {index = 0l @@ at; offset = $4 c; init = $5 c func} @@ at }
+  | LPAR ELEM bind_var_opt extern_kind elemindex_list RPAR
```
rossberg: Is it possible to factor out the choice between `extern_kind elemindex_list` and `elem_type elemref_list` into a separate production, thereby avoid all the multiplicative cases here?

interpreter/text/parser.mly
```diff
@@ -620,20 +636,20 @@ table_fields :
   | inline_export table_fields  /* Sugar */
     { fun c x at -> let tabs, elems, ims, exs = $2 c x at in
       tabs, elems, ims, $1 (TableExport x) c :: exs }
-  | elem_type LPAR ELEM active_elemref_list RPAR  /* Sugar */
+  | elem_type LPAR ELEM elemindex_list RPAR  /* Sugar */
```
rossberg: You should equally allow elemref_list in this shorthand now.

document/core/text/modules.rst
```diff
@@ -483,10 +483,21 @@ Element segments allow for an optional :ref:`table index <text-tableidx>` to ide
...
+     \text{(}~\text{elem}~~\Tid^?~~e{:}\Toffset_I~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
+       \Rightarrow\quad \{ \ETABLE~0, \EOFFSET~e, \EINIT~y^\ast \} \\ &&|&
+     \text{(}~\text{elem}~~\Tid^?~~\text{(}~\text{table}~~x{:}\Ttableidx_I ~\text{)}~~e{:}\Toffset_I~~\text{func}~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
        \Rightarrow\quad \{ \ETABLE~x, \EOFFSET~e, \EINIT~y^\ast \} \\ &&|&
-     \text{(}~\text{elem}~~\Tid^?~~et{:}\Telemtype~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{ \ETYPE~et, \EINIT~y^\ast \} \\
+     \text{(}~\text{elem}~~\Tid^?~~e{:}\Toffset_I~~y^\ast{:}\Tvec(\Texpr_I)~\text{)} \\ &&& \qquad
+       \Rightarrow\quad \{ \ETABLE~0, \EOFFSET~e, \EINIT~y^\ast \} \\ &&|&
+     \text{(}~\text{elem}~~\Tid^?~~\text{(}~\text{table}~~x{:}\Ttableidx_I ~\text{)}~~e{:}\Toffset_I~~et{:}\Telemtype~~y^\ast{:}\Tvec(\Texpr_I)~\text{)} \\ &&& \qquad
+       \Rightarrow\quad \{ \ETABLE~x, \EOFFSET~e, \ETYPE~et, \EINIT~y^\ast \} \\ &&|&
```
rossberg: AFAICT, I case like this is not currently supported by the elem AST as defined, so you'll need to change that as well, and all rules in the spec that use it.

See my comments in the parser as well. I think we should use the following abstract syntax (in both spec and impl):
```
<elem> ::=
  { type <elemtype>, init <expr>* }
  { type <elemtype>, init <expr>*, table <tableidx>, offset <expr> }
```
The former is passive, the latter active.

interpreter/syntax/ast.ml
rossberg: > The reason for this distinction is that it prevents an inconsistent state.

The elem_kind won't occur in the AST either, so there is no inconsistency. Simply think of the index forms as a shorthand in the binary format. The "desugaring" converts the elem_kind into an elem_type.

We want to keep the AST as simple and canonical as possible, so local optimisations of the binary format should not leak into it.

> About sharing code between element segments and data segments: It felt like the overlap was already limited

Yeah, but it's still nice to share it. Unless there are strong reasons for removing it I'd prefer not to.

> Question: why were there separate definitions for Active and Passive segments? Could they not also just be distinguished by a flag?

Well, the constructor is the flag. And the presence of some fields depends on it, which a Boolean flag couldn't express, not without allowing inconsistent state ;).

document/core/text/modules.rst
rossberg: The thing that's being defined here.

interpreter/binary/decode.ml
```diff
@@ -642,16 +630,46 @@ let passive_elem s =
...
+  | 2l ->
+    let index = at var s in
+    let offset = const s in
+    let _ = extern_kind s in
+    let init = elem_indices s in
+    ActiveWithIndices {index; offset; init}
+  | 4l ->
+    let index = Source.(0l @@ Source.no_region) in
+    let offset = const s in
+    let etype = elem_type s in
```
rossberg: This seems incorrect, there is no elem_type in this form.

interpreter/binary/encode.ml
```diff
@@ -477,39 +477,45 @@ let encode m =
...
-      segment active passive seg
+      match seg.it with
+      | ActiveWithIndices {index = {it = 0l;_}; offset; init} ->
+        u8 0x00; const offset; elem_indices init
+      | PassiveWithIndices {data} ->
+        u8 0x01; u8 0x00; elem_indices data
+      | ActiveWithIndices {index; offset; init} ->
+        u8 0x02; var index; const offset; u8 0x00; elem_indices init
+      | ActiveWithRefs {index = {it = 0l;_}; offset; etype; init} ->
+        u8 0x04; const offset; elem_type etype; vec elem_expr init
```
rossberg: Same here, there is no elem_type included in this case.

interpreter/text/parser.mly
rossberg: Yeah, it's a it annoying, but it follows from the independence of the flag bits in the flag word, see [here](https://github.com/WebAssembly/bulk-memory-operations/issues/98#issuecomment-507330729).

For the text format, I think we can factorise the legacy support slightly differently and do what you're asking, i.e., only allow omitting the elem_kind. Then you should be able to do the refactoring I suggested with only one additional special case for legacy active.

document/core/text/modules.rst
```diff
@@ -483,10 +493,20 @@ Element segments allow for an optional :ref:`table index <text-tableidx>` to ide
...
    \production{element segment} & \Telem_I &::=&
-     \text{(}~\text{elem}~~\Tid^?~~x{:}\Ttableidx_I~~\text{(}~\text{offset}~~e{:}\Texpr_I~\text{)}~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{ \ETABLE~x, \EOFFSET~e, \EINIT~y^\ast \} \\ &&|&
-     \text{(}~\text{elem}~~\Tid^?~~et{:}\Telemtype~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{ \ETYPE~et, \EINIT~y^\ast \} \\
+     \text{(}~\text{elem}~~\Tid^?~~e{:}\Toffset_I~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
+       \Rightarrow\quad \{ \ETABLE~0, \EOFFSET~e, \ETYPE~\FUNCREF, \EINIT~y^\ast \} \\ &&|&
+     \text{(}~\text{elem}~~\Tid^?~~e{:}\Toffset_I~~\{et, init\}{:}\Tfunclist_I~\text{)} \\ &&& \qquad
+       \Rightarrow\quad \{ \ETABLE~0, \EOFFSET~e, \ETYPE~et, \EINIT~init \} \\ &&|&
+     \text{(}~\text{elem}~~\Tid^?~~\text{(}~\text{table}~~x{:}\Ttableidx_I ~\text{)}~~e{:}\Toffset_I~~\{et, init\}{:}\Tfunclist_I~\text{)} \\ &&& \qquad
```
rossberg: Notation for pattern-matching on records via their punning on their labels, like here, hasn't been defined, so I suggest having funclist simple return a pair `(et,y*)` instead.

document/core/valid/modules.rst
```diff
@@ -147,7 +147,7 @@ Element Segments
 
 Element segments :math:`\elem` are classified by :ref:`segment types <syntax-segtype>`.
 
-:math:`\{ \ETABLE~x, \EOFFSET~\expr, \EINIT~e^\ast \}`
+:math:`\{ \ETABLE~x, \EOFFSET~\expr, \ETYPE~et, \EINIT~e^\ast \}`
 ......................................................
```
rossberg: Adjust underline, otherwise Sphinx will be unhappy.

document/core/valid/modules.rst
```diff
@@ -175,15 +177,19 @@ Element segments :math:`\elem` are classified by :ref:`segment types <syntax-seg
      \qquad
      C \vdashexprconst \expr \const
      \qquad
+     C \vdashexprconst et = \FUNCREF
```
rossberg: There is no C vdash here, since this is just a plain syntactic equality.

document/core/valid/modules.rst
```diff
@@ -193,6 +199,8 @@ Element segments :math:`\elem` are classified by :ref:`segment types <syntax-seg
 
 .. math::
    \frac{
+     C \vdashexprconst et = \FUNCREF
```
rossberg: Same here.

interpreter/binary/decode.ml
```diff
@@ -146,6 +146,11 @@ let elem_type s =
   | -0x10 -> FuncRefType
   | _ -> error s (pos s - 1) "invalid element type"
 
+let extern_kind s =
```
rossberg: Rename to elem_kind, move to Tables part.

document/core/syntax/modules.rst
```diff
@@ -261,7 +261,7 @@ The |MELEM| component of a module defines a vector of element segments. Each act
 .. math::
    \begin{array}{llll}
    \production{element segment} & \elem &::=&
-     \{ \ETABLE~\tableidx, \EOFFSET~\expr, \EINIT~\vec(\elemexpr) \} \\&&|&
+     \{ \ETABLE~\tableidx, \EOFFSET~\expr, \ETYPE~\elemtype, \EINIT~\vec(\elemexpr) \} \\&&|&
```
rossberg: Have you adjusted all the use sites of this record def?

interpreter/binary/decode.ml
```diff
@@ -613,27 +618,10 @@ let code_section s =
...
-    Passive {etype; data}
-  | 2l ->
-    let index = at var s in
-    let offset = const s in
-    let init = active s in
-    Active {index; offset; init}
-  | _ -> error s (pos s - 1) "invalid segment kind"
-
-let active_elem s =
+let func_index s =
```
rossberg: Rename to elem_index.

interpreter/binary/encode.ml
```diff
@@ -477,39 +477,48 @@ let encode m =
...
-      | RefFunc x -> u8 0xd2; var x; end_ ()
+    let elem_indices data = vec elem_index data
 
     let table_segment seg =
-      let active init = vec active_elem init in
-      let passive etype data = elem_type etype; vec passive_elem data in
-      segment active passive seg
+      match seg.it with
+      | ElemActive {index = {it = 0l;_}; offset; init; _}
+        when not (contains_null_ref init) ->
```
rossberg: We will add more expressions in the future, so a more forward-compatible predicate would be `all_func_ref`.

interpreter/binary/encode.ml
```diff
@@ -477,39 +477,48 @@ let encode m =
...
     (* Data section *)
     let memory_segment seg =
-      segment string (fun _ s -> string s) seg
+      match seg.it with
+      | DataActive {index = {it = 0l;_}; offset; init} ->
+          u8 0x00; const offset; string init
+      | DataPassive {data} ->
+          u8 0x01; string data
+      | DataActive {index; offset; init} ->
+          u8 0x02; var index; const offset; string init
```
rossberg: Nit: inconsistent indentation

interpreter/syntax/ast.ml
```diff
@@ -140,18 +140,21 @@ and memory' =
...
 and elem' =
   | RefNull
   | RefFunc of var
 
-type table_segment = (elem list, elem_type) segment
-type memory_segment = (string, unit) segment
+
+type table_segment = table_segment' Source.phrase
+and table_segment' =
+  | ElemActive of {index : var; offset : const; etype : elem_type; init : elem list}
```
rossberg: Nit: Switch name around: `ActiveElem`, same for others

interpreter/syntax/ast.ml
```diff
@@ -264,3 +267,6 @@ let string_of_name n =
   in
   List.iter escape n;
   Buffer.contents b
+
+let contains_null_ref l = List.exists (fun elem -> elem.it = RefNull) l
```
rossberg: For hygiene, I'd prefer not to dump random helper function in the AST module. Given how trivial the function is, it seems fine to duplicate it in the two modules that use it.

interpreter/text/arrange.ml
```diff
@@ -296,29 +296,44 @@ let memory off i mem =
...
+    Node ("elem", Node ("offset", const offset) :: atom elem_type etype
+    :: list elem_expr init)
+  | ElemActive {index; offset; etype; init} ->
+    Node ("elem", Node ("table", [atom var index])
+    :: Node ("offset", const offset)
+    :: atom elem_type etype :: list elem_expr init)
+  | ElemPassive {data; _}
+    when not (contains_null_ref data) ->
+    Node ("elem func", list elem_index data)
+  | ElemPassive {etype; data} -> Node ("elem", atom elem_type etype
```
rossberg: Nit: line break after `->`

interpreter/text/parser.mly
```diff
@@ -566,42 +566,58 @@ func_body :
 
 /* Tables, Memories & Globals */
 
+table_ref :
```
rossberg: Rename to `table_use`, in analogy to the existing `type_use`.

interpreter/text/parser.mly
```diff
@@ -566,42 +566,58 @@ func_body :
...
+  | LPAR TABLE var RPAR { fun c -> $3 c }
+
 offset :
   | LPAR OFFSET const_expr RPAR { $3 }
   | expr { let at = at () in fun c -> $1 c @@ at }  /* Sugar */
 
+elem_kind :
+  | FUNC { FuncRefType }
+
 elemref :
```
rossberg: While you're here, can you rename this to `elem_expr` (and `elem_expr_list` below)?

interpreter/text/parser.mly
```diff
@@ -566,42 +566,58 @@ func_body :
...
   | LPAR REF_FUNC var RPAR { let at = at () in fun c -> ref_func ($3 c func) @@ at }
 
-passive_elemref_list :
+elemref_list :
   | /* empty */ { fun c -> [] }
-  | elemref passive_elemref_list { fun c -> $1 c :: $2 c }
+  | elemref elemref_list { fun c -> $1 c :: $2 c }
 
-active_elemref_list :
+elemindex_list :
```
rossberg: Nit: `elem_var_list`.

document/core/text/modules.rst
```diff
@@ -483,10 +493,20 @@ Element segments allow for an optional :ref:`table index <text-tableidx>` to ide
 .. math::
    \begin{array}{llclll}
    \production{element segment} & \Telem_I &::=&
-     \text{(}~\text{elem}~~\Tid^?~~x{:}\Ttableidx_I~~\text{(}~\text{offset}~~e{:}\Texpr_I~\text{)}~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{ \ETABLE~x, \EOFFSET~e, \EINIT~y^\ast \} \\ &&|&
-     \text{(}~\text{elem}~~\Tid^?~~et{:}\Telemtype~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
-       \Rightarrow\quad \{ \ETYPE~et, \EINIT~y^\ast \} \\
+     \text{(}~\text{elem}~~\Tid^?~~e{:}\Toffset_I~~y^\ast{:}\Tvec(\Tfuncidx_I)~\text{)} \\ &&& \qquad
```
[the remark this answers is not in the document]
gahaas: Done.

document/core/syntax/modules.rst
gahaas: I checked it, yes. I don't know for sure though if I found all use sites.

document/core/valid/modules.rst
gahaas: Done.

document/core/valid/modules.rst
gahaas: Done.

document/core/valid/modules.rst
gahaas: Done.

interpreter/binary/decode.ml
gahaas: Done.

interpreter/binary/decode.ml
gahaas: Done.

interpreter/binary/encode.ml
gahaas: Done.

interpreter/binary/encode.ml
gahaas: Done.

interpreter/syntax/ast.ml
gahaas: Done.

interpreter/syntax/ast.ml
gahaas: Done.

interpreter/text/arrange.ml
gahaas: Done.

interpreter/text/parser.mly
gahaas: Done.

interpreter/text/parser.mly
gahaas: Done.

interpreter/text/parser.mly
gahaas: Done.

document/core/syntax/modules.rst
rossberg: Shouldn't this at least require a change to the corresponding binary format decoding rule as well?

gahaas: This was already done in https://github.com/WebAssembly/bulk-memory-operations/pull/108.

rossberg: Ah sorry, I lost track of the changes. So I guess we should already have changed the AST in that PR, but I missed that while reviewing.

Anyway, LGTM then.

PR #110, opened 2019-08-30
Fix the binary.wast test for multi-bytes table indices

gahaas:
One test in binary.wast checked if the table index can be encoded as varint. However, it encoded the table index not in the table index field but in the flag field. I fixed that now, and I also fixed the decoder to not accept varints as the flag field anymore.

### DIFF
--- a/interpreter/binary/decode.ml
+++ b/interpreter/binary/decode.ml
@@ -637,32 +637,32 @@ let elem_refs s =
   vec (at elem_expr) s
 
 let table_segment s =
-  match vu32 s with
-  | 0l ->
+  match u8 s with
+  | 0 ->
     let index = Source.(0l @@ Source.no_region) in
     let offset = const s in
     let init = elem_indices s in
     ActiveElem {index; offset; etype = FuncRefType; init}
-  | 1l ->
+  | 1 ->
     let etype = elem_kind s in
     let data = elem_indices s in
     PassiveElem {etype; data}
-  | 2l ->
+  | 2 ->
     let index = at var s in
     let offset = const s in
     let etype = elem_kind s in
     let init = elem_indices s in
     ActiveElem {index; offset; etype; init}
-  | 4l ->
+  | 4 ->
     let index = Source.(0l @@ Source.no_region) in
     let offset = const s in
     let init = elem_refs s in
     ActiveElem {index; offset; etype = FuncRefType; init}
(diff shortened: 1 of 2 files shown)


interpreter/binary/decode.ml
```diff
@@ -637,32 +637,32 @@ let elem_refs s =
   vec (at elem_expr) s
 
 let table_segment s =
-  match vu32 s with
-  | 0l ->
+  match u8 s with
+  | 0 ->
```
rossberg: Stylistic nit: I typically write `0xXY` to express direct byte values.

test/core/binary.wast
```diff
@@ -129,9 +129,9 @@
   "\00asm" "\01\00\00\00"
   "\04\04\01"                          ;; Table section with 1 entry
   "\70\00\00"                          ;; no max, minimum 0, funcref
-  "\09\07\01"                          ;; Element section with 1 entry
-  "\80\00"                             ;; Table index 0, encoded with 2 bytes
-  "\41\00\0b\00"                       ;; (i32.const 0) with no elements
```
sbc100: I'm having a little trouble integrating this in wabt.     Why was an extra byte added to the end here?    What does it represent.   The `\b` terminates the init expt and the a single zero represents the num of elements, no?

rossberg: Note that this is segment format `02` now, which requires an elem kind. That is the first 0 byte after the init expr.

sbc100: I see.  Sorry I think I missed that change.  

Is the documentation out-to-date, or maybe I'm missing something?   From the overview.md and from published spec it looks like the what follows the init_expr is just a list of `func_index`:
https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md#element-segments
https://webassembly.github.io/bulk-memory-operations/core/binary/modules.html#binary-elemsec

rossberg:
Hm, yes, indeed, seems like nobody ever fixed the overview; spec and interpreter were changed in #108 and #109, but apparently travis hasn't been set up for this repo, so gh.io hasn't been updated since July.

lars-t-hansen:
> Hm, yes, indeed, seems like nobody ever fixed the overview.

That's #116, I was planning to get to it this week but anyone should feel free to grab it :)

Issue #111, opened 2019-09-12
memory.copy|fill semantics limit optimizations for short constant lengths


eqrion:
As discussed in #1, the expectation is that producers will use `memory.copy|fill` for short lengths in addition to long lengths. We’ve already seen this to be the case, and have been investigating a performance regression resulting from LLVM 9 using `memory.copy` for short constant lengths [1].

Part of that regression is in a sub-optimal OOL call to the system `memmove`, but to really get performance to par, we’d like to inline these short `memory.copy`s to loads and stores.

This has turned out challenging to implement in a way that is better or equal to the Wasm loads and stores that were emitted previously.

There are several problems resulting from the following:

1. We must assume no alignment for `src`, `dest`
2. When a range is partially OOBs, we must write all bytes up until the region is OOBs
3. We must copy correctly when there is overlap of `src`, `dest`

(1) and (2) are related. Because we don’t know the alignment of `src` or `dest` we cannot use wider transfers than a single byte at a time (e.g 32bit, 64bit, or 128bit) or else we’d be at risk of the final store being partially OOB and not writing all bytes up to the boundary due to misalignment of `src` or `dest`.

The system `memmove` can work around this by aligning the `src`/`dest` pointers, using wide transfer widths, and fixing up slop afterwards. But this isn’t feasible for inlined code.

The problem with (3) is that we need to generate two sequences of loads and stores. One for if `src < dest` where we need to copy from `high -> low`, and another for `low -> high`. This adds to code size and is a comparison that we didn’t need to do before. This could be potentially solved in a branchless way by using vector registers as a temporary buffer, but that has difficulty still with (1) and (2).

There seem to be several options that could improve this situation.

1. We could ask LLVM to not emit `memory.copy` in these situations. `memory.copy` is not equivalent to `memcpy`, it’s `memmove` and has more strict semantics than LLVM requires. For example, with struct copies LLVM should know the alignment and that there is no overlap. Recovering this information at runtime is unfortunate. The downside to this is potential binary size increases, and limiting to loads and store widths that are defined in Wasm (e.g. no SIMD yet).

2. We could modify the behavior for partially OOB’s ranges to not write any bytes at all. This would allow us to load all `src` bytes into vector registers, then store them to `dest` from `high -> low`. If there is a trap, it will happen immediately and nothing will be written. This fixes (1) and (3) by changing (2). The uncertainty here is around whether this is possible with a future ‘memory.protect’ instruction.

[...]

PR #123, opened 2019-11-05
[exec] Bounds check bulk-memory before execution

eqrion:
Spec issue: https://github.com/webassembly/bulk-memory-operations/issues/111

This commit changes the semantics of bulk-memory instructions to perform
an upfront bounds check and trap if any access would be out-of-bounds without
writing.

This affects the following:
 * memory.init/copy/fill
 * table.init/copy (fill requires reftypes)
 * data segment init (lowers to memory.init)
 * elem segment init (lowers to table.init)

This commit corrects an edge case where a memory/table.init with zero
length wouldn't trap even if the data/elem segment had been dropped.
This was due to an extra pattern in eval that would match zero-length
instructions and not check if the segment had been dropped. One test
case in bulk.wast had to be changed for this.


eqrion:
I would also be interested in updating the spec-text as well, but will need some guidance.

Right now the instructions are implemented by lowering to plain i32.load8_u/i32.store8's for memory and table.get/table.set for tables.

I think there are a couple options here.

1. Use the stack as a temporary buffer.
   * Lower to a series of loads onto the stack, followed by a sequence of stores from high addresses to low.
1. Add new administrative functions.
   * transfer_memory(src: vec(u8), memory: mem_address, dest: offset)
   * transfer_elements(src: vec(elem), table: tab_address, dest: offset)
   * Lower init,copy,fill to these by constructing the appropriate src buffer.
1. Add a bounds check at the beginning of the existing semantics, keep everything else.

I think all of these can be equivalent in the single-threaded case, but I'm having a hard time reasoning about the implications for threading.

rossberg:
Hi @eqrion, thanks for moving forward with this. Turns out we were in a race. :)  I had started on a similar patch last week but didn't get to finish the spec part yet. I uploaded it [here](https://github.com/WebAssembly/bulk-memory-operations/compare/pre-oob) so that we can compare. But mine is more WIP. In particular, I did not properly fix the test generators. And I only started on the spec itself. So we should probably go with yours instead.

However, one relevant difference is that I kept and adapted the small-step semantics for the interpreter instead of going back to monolithic implementations. This is important for two reasons: (1) the interpreter should match the spec rules, and (2) the rules need to use small-step in order to be forward-compatible with the threads proposal, where bulk ops are no longer atomic operations.

How would you feel about incorporating my interpreter implementation into your patch?

eqrion:
> Hi @eqrion, thanks for moving forward with this. Turns out we were in a race. :) I had started on a similar patch last week but didn't get to finish the spec part yet. I uploaded it [here](https://github.com/WebAssembly/bulk-memory-operations/compare/pre-oob) so that we can compare. But mine is more WIP. In particular, I did not properly fix the test generators. And I only started on the spec itself. So we should probably go with yours instead.
> 
> However, one relevant difference is that I kept and adapted the small-step semantics for the interpreter instead of going back to monolithic implementations. This is important for two reasons: (1) the interpreter should match the spec rules, and (2) the rules need to use small-step in order to be forward-compatible with the threads proposal, where bulk ops are no longer atomic operations.
> 
> How would you feel about incorporating my interpreter implementation into your patch?

Ah apologies about the duplicated effort, I should have posted that I was taking a look at this. :)

Good to know about needing the small-step semantics, I think that was what I was getting at with my question. I'll adapt your interpreter implementation into my patch.

rossberg:
Cool, thanks!

Forgot one thing. You mentioned:
> This commit corrects an edge case where a memory/table.init with zero
length wouldn't trap even if the data/elem segment had been dropped.

This actually was an intentional fix in #115. The rationale being that accessing a dropped segment is equivalent to an out-of-bounds access, because a dropped segment is essentially a segment shrunk to size 0. A zero-length access should therefore behave consistently with no effect.

eqrion:
> Forgot one thing. You mentioned:
> 
> > This commit corrects an edge case where a memory/table.init with zero
> > length wouldn't trap even if the data/elem segment had been dropped.
> 
> This actually was an intentional fix in #115. The rationale being that accessing a dropped segment is equivalent to an out-of-bounds access, because a dropped segment is essentially a segment shrunk to size 0. A zero-length access should therefore behave consistently with no effect.

Ah, I believe the [exec-text](https://github.com/WebAssembly/bulk-memory-operations/blob/master/document/core/exec/instructions.rst#memoryinitx) for `memory.init` needs an update then. Step 10 does a trap if the data segment has been dropped, before step 17 does the check for `cnt = 0`. That was my reasoning for calling this a bug.

Spidermonkey will also need an update for this, as we will trap in this case.

eqrion:
> Ah, I believe the [exec-text](https://github.com/WebAssembly/bulk-memory-operations/blob/master/document/core/exec/instructions.rst#memoryinitx) for `memory.init` needs an update then. Step 10 does a trap if the data segment has been dropped, before step 17 does the check for `cnt = 0`. That was my reasoning for calling this a bug.

On further examination, it appears that the formal rules are correct and the prose is what may need an update.

rossberg:
Ah, I see. If you want, please feel free to fix this as part of this PR.

eqrion:
I've updated the interpreter commit to use your patch, keep the small step semantics, and extend early bounds checking to the table operations. That commit should be good to go.

I've also added a commit for the dropped segments issue, and started looking at the spec text.

document/core/exec/instructions.rst
```diff
@@ -689,24 +689,21 @@ Memory Instructions
    S; F; (\I32.\CONST~i)~\val~(\I32.\CONST~0)~(\MEMORYFILL) &\stepto& S; F; \epsilon
    \end{array} \\
    \begin{array}{lcl@{\qquad}l}
-   S; F; (\I32.\CONST~i)~\val~(\I32.\CONST~1)~(\MEMORYFILL) &\stepto& S; F;
-     (\I32.\CONST~i)~\val~(\I32\K{.}\STORE\K{8}~\{ \OFFSET~0, \ALIGN~0 \}) \\
-   \end{array} \\
+   S; F; (\I32.\CONST~i)~\val~(\I32.\CONST~n)~(\MEMORYFILL) &\stepto& S; F; \TRAP \\
```
rossberg: All the spec changes from my patch are still very incomplete and WIP, so I'd rather leave them out of this PR.

test/core/bulk.wast
```diff
@@ -128,14 +128,14 @@
...
-;; Out-of-bounds writes trap, but all previous writes succeed.
+;; Out-of-bounds writes trap, and nothing is written.
 (assert_trap (invoke "init" (i32.const 0xfffe) (i32.const 0) (i32.const 3))
     "out of bounds memory access")
-(assert_return (invoke "load8_u" (i32.const 0xfffe)) (i32.const 0xaa))
-(assert_return (invoke "load8_u" (i32.const 0xffff)) (i32.const 0xbb))
+(assert_return (invoke "load8_u" (i32.const 0xfffe)) (i32.const 0x00))
+(assert_return (invoke "load8_u" (i32.const 0xffff)) (i32.const 0x00))
+
+;; Init ending at memory limit and segment limit is ok.
```
rossberg: May be preferable to leave this where it was before and then tweak the two asserts above to check for 0xcc and 0xdd?

interpreter/exec/eval.ml
```diff
@@ -117,9 +117,25 @@ let drop n (vs : 'a stack) at =
...
-  let k = I32.add i j in
-  if I32.lt_u k i then Trapping msg else Plain (Const (I32 k @@ at))
+let mem_oob frame x i n =
+  I64.gt_u (I64.add (I64_convert.extend_i32_u i) (I64_convert.extend_i32_u n))
+    (Memory.bound (memory frame.inst x))
+
+let data_oob frame x i n =
+  match !(data frame.inst x) with
+    | None -> false
+    | Some bs -> I64.gt_u (I64.add (I64_convert.extend_i32_u i) (I64_convert.extend_i32_u n))
```
rossberg: Nit: line break after the `->`

interpreter/exec/eval.ml
```diff
@@ -117,9 +117,25 @@ let drop n (vs : 'a stack) at =
...
+      (I64.of_int_u (String.length bs))
+
+let table_oob frame x i n =
+  I64.gt_u (I64.add (I64_convert.extend_i32_u i) (I64_convert.extend_i32_u n))
+    (I64_convert.extend_i32_u (Table.size (table frame.inst x)))
+
+let elem_oob frame x i n =
+  match !(elem frame.inst x) with
+    | None -> false
+    | Some es -> I64.gt_u (I64.add (I64_convert.extend_i32_u i) (I64_convert.extend_i32_u n))
```
rossberg: Same here.

interpreter/exec/eval.ml
```diff
@@ -291,71 +311,67 @@ let rec step (c : config) : config =
...
 
       | MemoryInit x, I32 0l :: I32 s :: I32 d :: vs' ->
         vs', []
 
-      | MemoryInit x, I32 1l :: I32 s :: I32 d :: vs' ->
+      | MemoryInit x, I32 n :: I32 s :: I32 d :: vs'
+        when mem_oob frame (0l @@ e.at) d n ->
+        vs', [Trapping (memory_error e.at Memory.Bounds) @@ e.at]
+
+      | MemoryInit x, I32 n :: I32 s :: I32 d :: vs'
```
rossberg: For symmetry, merge this with the previous case, as you did for TableInit.

eqrion:
@rossberg Is this ready to go now?

eqrion:
Thanks!

Issue #124, opened 2019-11-08
Zero-sized memory.init of dropped segments should trap


tlively:
I'm currently working on generalizing Binaryen's memory-packing pass to optimize passive segments in addition to the active segments it already optimizes. The main idea is that when there is a long run of zeroes in a data segment, that data segment is split into two segments that do not contains those zeroes and any `memory.init` instructions that reference that segment are split into a sequence of `memory.init` and `memory.fill` instructions.

The problem is that if a `memory.init` instruction uses a section of a data segment that is only comprised of zeroes, it naively would be replaced with a single `memory.fill` instruction. But that would not be a correct transformation because the `memory.fill` instruction would not trap if the original data segment would have been dropped. Emulating the trapping behavior correctly would involve creating new globals to track the hidden state of each of the data segments that are split, which is unnecessarily complex. If zero-sized `memory.init` instructions could trap, a simpler solution would be to insert a zero-size `memory.init` of one of the split data segments before the `memory.fill`.

My understanding is that zero-length `memory.init` currently does not trap as a matter of convenience, so would anyone object to changing this behavior?

rossberg:
The rationale for the current semantics is that a dropped segment behaves like a zero-sized one and both failure conditions are essentially the same error category. Arguably, that is the simplest and most natural semantics, and jits to the simplest code. In fact, I was about to propose that we treat drop *exactly* like a shrink-to-zero, because that would get rid of having to track the dropped state as a separate piece of information, which is kind of annoying atm. (The only observable difference would be that dropping a segment twice would be a nop instead of a trap.)

So, hm, what you propose would make this simplification impossible.

Could you solve your use case by always keeping at least one byte for memory.fill? That may be slightly less elegant, but in practice, is it much different from doing a zero-sized fill?

Another question is why you even care about maintaining the error. The trap is a fatal failure, i.e., the change in semantics only affects broken programs. Does that matter for your use case?

Tangential question: Why only replace runs of zeroes? Wouldn't the same optimisation apply for any byte value?

tlively:
> Could you solve your use case by always keeping at least one byte for memory.fill? That may be slightly less elegant, but in practice, is it much different from doing a zero-sized fill?

[...]

PR #126, opened 2019-11-21
[interpreter] Simplify zero-len and drop semantics

rossberg:
As discussed on #124, this introduces two simplifications, partially as a consequence of #111/#123:

1. Zero-length out-of-bound access trap again.

   This reverts #100, #101, #102, which turn out to no longer be simplifications but rather complications in the presence of the OOB semantics more recently introduced by #111/#123.

2. Segment drops are treated like a shrink-to-zero.

   This unifies drop check and OOB into a single check and avoids maintaining dropped as an extra state bit in the implementation. The only observable change in behaviour is that it is no longer a trap to (a) drop a segment twice, or (b) perform a 0-length init from a dropped segment at offset 0; both these cases are nops now.

Taken together, these changes also address #124.

Includes:

- Overview changes.
- Interpreter changes.
- Test changes (generated tests to be added by @eqrion).
- Spec of drop semantics.

The spec of OOB checks follows in PR #129, since resp changes for #123 were still missing.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -728,7 +728,7 @@ Memory Instructions
 
 8. Assert: due to :ref:`validation <valid-memory.init>`, :math:`S.\SDATA[\X{da}]` exists.
 
-9. Let :math:`\X{data}^?` be the optional :ref:`data instance <syntax-datainst>` :math:`S.\SDATA[\X{da}]`.
+9. Let :math:`\X{data}` be the  :ref:`data instance <syntax-datainst>` :math:`S.\SDATA[\X{da}]`.
 
 10. Assert: due to :ref:`validation <valid-memory.init>`, a value of :ref:`value type <syntax-valtype>` |I32| is on the top of the stack.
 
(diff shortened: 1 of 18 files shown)


lars-t-hansen:
Punting to @eqrion for work on tests :-)

eqrion:
I wrote a [commit](https://github.com/eqrion/wasm-bulk-memory-operations/commit/438b85d8cdb21882bba3e66b6dd0146faad16b6c) based on your branch to update the test-generators. I can confirm that it resolved all of the test failures.

Feel free to take it and use in this PR, I'm unsure of the best way on GH to handle this situation.

rossberg:
@eqrion, thanks! Can't you simply create a PR against this branch?

rossberg:
With @eqrion's test generator tweaks, this PR should be complete and ready for review now.

rossberg:
@eqrion, GH doesn't let me request you as a reviewer, but can you please take a look?

eqrion:
> @eqrion, thanks! Can't you simply create a PR against this branch?

Ah, this is a branch in this repo not a personal one. I suppose it would work either way, though. I just opened the PR.

> @eqrion, GH doesn't let me request you as a reviewer, but can you please take a look?

Yup, I'll take a look at this today.

Issue #128, opened 2019-11-21
[spec] Extend store typing in appendix with elem & data instances


rossberg:

rossberg:
Fixed by #129.

Issue #127, opened 2019-11-21
[spec] Make OOB spec changes to follow #123 and #126


rossberg:
Interpreter and tests have been adapted for new OOB early checking behaviour in #123 and #126, but the respective spec changes still have to be done.

rossberg:
Fixed by #129.

PR #129, opened 2019-11-21
[spec] Adapt OOB behaviour in spec; store typing

rossberg:
This does the necessary spec changes to address #127 and #128:

- Spec new OOB behaviour with early checks and trapping on size 0 (again).
- Extend store typing and store extension to element and data instances.
- Some streamlining.

Baseline is #126; all changes outside spec dir are from that PR.

Any volunteers for reviewing?

### DIFF
--- a/document/core/appendix/index-rules.rst
+++ b/document/core/appendix/index-rules.rst
@@ -54,6 +54,8 @@ Construct                                        Judgement
 :ref:`Table instance <valid-tableinst>`          :math:`S \vdashtableinst \tableinst : \tabletype`
 :ref:`Memory instance <valid-meminst>`           :math:`S \vdashmeminst \meminst : \memtype`
 :ref:`Global instance <valid-globalinst>`        :math:`S \vdashglobalinst \globalinst : \globaltype`
+:ref:`Element instance <valid-eleminst>`         :math:`S \vdasheleminst \eleminst \ok`
+:ref:`Data instance <valid-datainst>`            :math:`S \vdashdatainst \datainst \ok`
 :ref:`Export instance <valid-exportinst>`        :math:`S \vdashexportinst \exportinst \ok`
 :ref:`Module instance <valid-moduleinst>`        :math:`S \vdashmoduleinst \moduleinst : C`
 :ref:`Store <valid-store>`                       :math:`\vdashstore \store \ok`
@@ -95,6 +97,8 @@ Construct                                        Judgement
 :ref:`Table instance <extend-tableinst>`         :math:`\vdashtableinstextends \tableinst_1 \extendsto \tableinst_2`
 :ref:`Memory instance <extend-meminst>`          :math:`\vdashmeminstextends \meminst_1 \extendsto \meminst_2`
 :ref:`Global instance <extend-globalinst>`       :math:`\vdashglobalinstextends \globalinst_1 \extendsto \globalinst_2`
+:ref:`Elment instance <extend-eleminst>`         :math:`\vdasheleminstextends \eleminst_1 \extendsto \datainst_2`
+:ref:`Data instance <extend-datainst>`           :math:`\vdashdatainstextends \datainst_1 \extendsto \eleminst_2`
 :ref:`Store <extend-store>`                      :math:`\vdashstoreextends \store_1 \extendsto \store_2`
 ===============================================  ===============================================================================
 
--- a/document/core/appendix/index-rules.rst
+++ b/document/core/appendix/index-rules.rst
@@ -97,8 +97,8 @@ Construct                                        Judgement
 :ref:`Table instance <extend-tableinst>`         :math:`\vdashtableinstextends \tableinst_1 \extendsto \tableinst_2`
 :ref:`Memory instance <extend-meminst>`          :math:`\vdashmeminstextends \meminst_1 \extendsto \meminst_2`
 :ref:`Global instance <extend-globalinst>`       :math:`\vdashglobalinstextends \globalinst_1 \extendsto \globalinst_2`
-:ref:`Elment instance <extend-eleminst>`         :math:`\vdasheleminstextends \eleminst_1 \extendsto \datainst_2`
-:ref:`Data instance <extend-datainst>`           :math:`\vdashdatainstextends \datainst_1 \extendsto \eleminst_2`
+:ref:`Element instance <extend-eleminst>`         :math:`\vdasheleminstextends \eleminst_1 \extendsto \eleminst_2`
+:ref:`Data instance <extend-datainst>`           :math:`\vdashdatainstextends \datainst_1 \extendsto \datainst_2`
 :ref:`Store <extend-store>`                      :math:`\vdashstoreextends \store_1 \extendsto \store_2`
 ===============================================  ===============================================================================
 
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -716,8 +716,6 @@ Memory Instructions
 
 5. Let :math:`\X{mem}` be the :ref:`memory instance <syntax-meminst>` :math:`S.\SMEMS[\X{ma}]`.
 
-1. Let :math:`F` be the :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>`.
-
 6. Assert: due to :ref:`validation <valid-memory.init>`, :math:`F.\AMODULE.\MIDATAS[x]` exists.
 
 7. Let :math:`\X{da}` be the :ref:`data address <syntax-dataaddr>` :math:`F.\AMODULE.\MIDATAS[x]`.
@@ -762,9 +760,9 @@ Memory Instructions
 
 25. Push the value :math:`\I32.\CONST~(s+1)` to the stack.
 
-24. Push the value :math:`\I32.\CONST~(n-1)` to the stack.
+26. Push the value :math:`\I32.\CONST~(n-1)` to the stack.
 
-25. Execute the instruction :math:`\MEMORYINIT~x`.
+27. Execute the instruction :math:`\MEMORYINIT~x`.
 
 .. math::
    ~\\[-1ex]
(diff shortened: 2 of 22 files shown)


document/core/appendix/index-rules.rst
```diff
@@ -95,6 +97,8 @@ Construct                                        Judgement
 :ref:`Table instance <extend-tableinst>`         :math:`\vdashtableinstextends \tableinst_1 \extendsto \tableinst_2`
 :ref:`Memory instance <extend-meminst>`          :math:`\vdashmeminstextends \meminst_1 \extendsto \meminst_2`
 :ref:`Global instance <extend-globalinst>`       :math:`\vdashglobalinstextends \globalinst_1 \extendsto \globalinst_2`
+:ref:`Elment instance <extend-eleminst>`         :math:`\vdasheleminstextends \eleminst_1 \extendsto \datainst_2`
```
eqrion: Element

document/core/exec/instructions.rst
```diff
@@ -684,28 +678,28 @@ Memory Instructions
...
+     \quad\stepto\quad S; F; \TRAP
+     \\ \qquad
+     (\iff d + n > |S.\SMEMS[F.\AMODULE.\MIMEMS[x]].\MIDATA|) \\
+   \\[1ex]
+   S; F; (\I32.\CONST~d)~\val~(\I32.\CONST~0)~\MEMORYFILL
+     \quad\stepto\quad S; F; \epsilon
+     \\ \qquad
+     (\otherwise)
+   \\[1ex]
+   S; F; (\I32.\CONST~d)~\val~(\I32.\CONST~n+1)~\MEMORYFILL
```
eqrion: Out of curiosity, why the choice of 'n+1' and 'n' here? As opposed to 'n' and 'n-1'.

document/core/exec/instructions.rst
```diff
@@ -722,13 +716,15 @@ Memory Instructions
 
 5. Let :math:`\X{mem}` be the :ref:`memory instance <syntax-meminst>` :math:`S.\SMEMS[\X{ma}]`.
 
+1. Let :math:`F` be the :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>`.
```
eqrion: I believe this should be dropped.

document/core/exec/instructions.rst
```diff
@@ -742,78 +738,58 @@ Memory Instructions
...
+23. Push the value :math:`\I32.\CONST~(d+1)` to the stack.
 
-23. Push the value :math:`\vconst_{\I32}(dst+1)` to the stack.
+24. Assert: due to the earlier check against the memory size, :math:`s+1 < 2^{32}`.
 
-24. Push the value :math:`\vconst_{\I32}(src+1)` to the stack.
+25. Push the value :math:`\I32.\CONST~(s+1)` to the stack.
 
-25. Push the value :math:`\I32.\CONST~(cnt-1)` to the stack.
+24. Push the value :math:`\I32.\CONST~(n-1)` to the stack.
```
eqrion: Numbering needs to be corrected here.

document/core/exec/instructions.rst
```diff
@@ -863,101 +824,111 @@ Memory Instructions
...
 
-   c. Execute the instruction :math:`\I32\K{.}\LOAD\K{8\_u}~\{ \OFFSET~0, \ALIGN~0 \}`.
+13. If :math:`n = 0`, then:
 
-   d. Execute the instruction :math:`\I32\K{.}\STORE\K{8}~\{ \OFFSET~0, \ALIGN~0 \}`.
+   a. Return.
+
+14. If :math:`d \leq s`, then:
+
+   b. Push the value :math:`\I32.\CONST~d` to the stack.
```
eqrion: I believe you meant to load from the source offset here.

The lettering also needs to be corrected.

document/core/exec/instructions.rst
```diff
@@ -863,101 +824,111 @@ Memory Instructions
...
+
+   c. Execute the instruction :math:`\I32\K{.}\LOAD\K{8\_u}~\{ \OFFSET~0, \ALIGN~0 \}`.
 
-   e. Return.
+   e. Push the value :math:`\I32.\CONST~s` to the stack.
```
eqrion: If I recall correctly, store pops the value first then pops the offset. So the dest offset needs to be pushed before the load pushes the value.

document/core/exec/instructions.rst
```diff
@@ -863,101 +824,111 @@ Memory Instructions
...
+   b. Push the value :math:`\I32.\CONST~(d+n-1)` to the stack.
 
-   a. Push the value :math:`\vconst_{\I32}(dst+cnt-1)` to the stack.
+   c. Execute the instruction :math:`\I32\K{.}\LOAD\K{8\_u}~\{ \OFFSET~0, \ALIGN~0 \}`.
 
-   b. Push the value :math:`\vconst_{\I32}(src+cnt-1)` to the stack.
+   d. Assert: due to the earlier check against the memory size, :math:`s+n-1 < 2^{32}`.
 
-   c. Push the value :math:`\I32.\CONST~1` to the stack.
+   e. Push the value :math:`\I32.\CONST~(s+n-1)` to the stack.
```
eqrion: Ditto the store stack argument ordering issue from above.

document/core/exec/instructions.rst
```diff
@@ -974,97 +945,111 @@ Table Instructions
...
+14. If :math:`d \leq s`, then:
+
+   a. Push the value :math:`\I32.\CONST~d` to the stack.
+
+   b. Execute the instruction :math:`\TABLEGET`.
+
+   c. Push the value :math:`\I32.\CONST~s` to the stack.
 
-   a. Push the value :math:`\I32.\CONST~dst` to the stack.
+   d. Execute the instruction :math:`\TABLESET`.
```
eqrion: I don't think `table.set` has a full definition in this spec yet, but if it will match reftypes then we need to push the destination offset first. Like with `i32.store`.

document/core/exec/instructions.rst
```diff
@@ -974,97 +945,111 @@ Table Instructions
...
 
-   e. Return.
+13. If :math:`n = 0`, then:
 
-9. If :math:`dst <= src`, then:
+   a. Return.
+
+14. If :math:`d \leq s`, then:
+
+   a. Push the value :math:`\I32.\CONST~d` to the stack.
```
eqrion: I believe you meant to load from the source offset here.

document/core/exec/instructions.rst
```diff
@@ -1087,94 +1072,74 @@ Table Instructions
...
+23. Push the value :math:`\I32.\CONST~(d+1)` to the stack.
 
-23. Push the value :math:`\vconst_{\I32}(dst+1)` to the stack.
+24. Assert: due to the earlier check against the segment size, :math:`s+1 < 2^{32}`.
 
-24. Push the value :math:`\vconst_{\I32}(src+1)` to the stack.
+25. Push the value :math:`\I32.\CONST~(s+1)` to the stack.
 
-25. Push the value :math:`\I32.\CONST~(cnt-1)` to the stack.
+24. Push the value :math:`\I32.\CONST~(n-1)` to the stack.
```
eqrion: Numbering needs to be corrected here.

document/core/exec/instructions.rst
```diff
@@ -1087,94 +1072,74 @@ Table Instructions
...
+    a. Return.
 
-    e. Execute the instruction :math:`\TABLESET`.
+18. Let :math:`\funcelem` be the :ref:`function element <syntax-funcelem>` :math:`\X{elem}.\EIELEM[s]`.
 
-    f. Return.
+19. Push the value :math:`\I32.\CONST~d` to the stack.
 
-19. Push the value :math:`\I32.\CONST~dst` to the stack.
+20. Push the value :math:`\funcelem` to the stack.
```
eqrion: It may be more clear to reorder to 18, 20, then 19 so that the use of the funcelem value is right next to the definition.

Additionally, it's a bit odd to use funcelem here as a value, but I understand that's due to reftypes being a separate proposal. Is the intention for this to be an administrative instruction like table.get/set in the mean time?

document/core/exec/instructions.rst
rossberg: Ah, because that avoids having to put `n > 0` as an extra side condition and it makes the inductive nature of the definition more apparent.

document/core/exec/instructions.rst
rossberg: Done.

document/core/exec/instructions.rst
rossberg: Done.

document/core/exec/instructions.rst
rossberg: Done.

document/core/exec/instructions.rst
rossberg: Done.

document/core/exec/instructions.rst
rossberg: Done.

document/core/exec/instructions.rst
rossberg: Done.

document/core/exec/instructions.rst
rossberg: It actually has a temporary one below as and administrative instruction. But yes, this is superceded by the reftype proposal.

document/core/exec/instructions.rst
rossberg: Yes, exactly. This is all a temporary hack. We screwed up the dependencies for these proposals.

document/core/exec/instructions.rst
rossberg: Done.

rossberg:
Thanks for that thorough review!

document/core/exec/instructions.rst
eqrion: Oh nice, I missed there was an entry in execution for it. I took a look at it and it does first pop funcelem, then the offset. So we need to push the dest offset before source offset here, like with memory.copy. This needs a change in the 'else' branch and the formal semantics as well.

eqrion:
I took another look and left one more comment. After that, this and the interpreter change should be ready to go!

document/core/exec/instructions.rst
rossberg: Indeed. Done!

Issue #116, opened 2019-10-03
Overview has stale information about encoding


lars-t-hansen:
The Overview was never updated to correspond to changes made in #98.  I'll try to rectify.

lars-t-hansen:
(Working on this now.)

binji:
Looks like this can be closed now. Thanks, Lars!

PR #130, opened 2019-11-21
New segment encoding

lars-t-hansen:
Resolve TODOs and other outstanding issues; update the overview to the new encoding spec; add forward references to reference-types and multi-memory proposals where appropriate.

Pertains to #116.

### DIFF
--- a/proposals/bulk-memory-operations/Overview.md
+++ b/proposals/bulk-memory-operations/Overview.md
@@ -228,7 +228,7 @@ The meaning of the bits of the flag field for element segments is:
 | 0   | 0=is active, 1=is passive                        |
 | 1   | if bit 0 clear: 0=table 0, 1=has table index     | 
 |     | if bit 0 set: 0=active, 1=declared               |
-| 2   | 0=carries function indicies; 1=carries elemexprs |
+| 2   | 0=carries indicies; 1=carries elemexprs          |
 
 which yields this view, with the fields carried by each flag value:
 
--- a/proposals/bulk-memory-operations/Overview.md
+++ b/proposals/bulk-memory-operations/Overview.md
@@ -45,11 +45,11 @@ the following contents:
 
 ```wasm
 (func (param $dst i32) (param $src i32) (param $size i32) (result i32)
-  get_local $dst
-  get_local $src
-  get_local $size
+  local.get $dst
+  local.get $src
+  local.get $size
   memory.copy
-  get_local $dst)
+  local.get $dst)
 ```
 
 Here are the results on my machine (x86_64, 2.9GHz, L1 32k, L2 256k, L3 256k):
@@ -160,8 +160,6 @@ Filling a memory region can be accomplished with `memory.fill`:
 
 * `memory.fill`: fill a region of linear memory with a given byte value
 
-TODO: should we provide `memory.clear` and `table.clear` instead?
-
 The [binary format for the data
 section](https://webassembly.github.io/spec/core/binary/modules.html#binary-datasec)
 currently has a collection of segments, each of which has a memory index, an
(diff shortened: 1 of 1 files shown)


proposals/bulk-memory-operations/Overview.md
```diff
@@ -171,93 +169,109 @@ Since WebAssembly currently does not allow for multiple memories, the memory
...
-The new binary format for element segments is similar to the new format for data segments, but
-also includes an element type when the segment is passive. A passive segment also has a sequence
-of `expr`s instead of function indices.
+The meaning of the bits of the flag field for element segments is:
+
+| Bit | Meaning                                          |
+| -   | -                                                |
+| 0   | 0=is active, 1=is passive                        |
+| 1   | if bit 0 clear: 0=table 0, 1=has table index     | 
+|     | if bit 0 set: 0=active, 1=declared               |
```
rossberg: You may want to add a note regarding `declared` since that actually isn't part of this proposal.

proposals/bulk-memory-operations/Overview.md
```diff
@@ -171,93 +169,109 @@ Since WebAssembly currently does not allow for multiple memories, the memory
...
-| 0 | Active | Always 0 | Present |
-| 1 | Passive | - | -  |
-| 2 | Active | Present | Present |
 
 ### Element segments
 
-The new binary format for element segments is similar to the new format for data segments, but
-also includes an element type when the segment is passive. A passive segment also has a sequence
-of `expr`s instead of function indices.
+The meaning of the bits of the flag field for element segments is:
```
rossberg: Maybe clarify that the flag field is a vu32.

proposals/bulk-memory-operations/Overview.md
```diff
@@ -272,43 +286,24 @@ The instruction has the signature `[i32 i32 i32] -> []`. The parameters are, in
 
 It is a validation error to use `memory.init` with an out-of-bounds segment index.
 
-A trap occurs if:
-* the segment is used after it has been dropped via `data.drop`. This includes
-  active segments that were dropped after being copied into memory during module
-  instantiation.
-* any of the accessed bytes lies outside the source data segment or the target memory
+A trap occurs if any of the accessed bytes lies outside the source data segment
```
rossberg: He, this will conflict with my PR. :)

proposals/bulk-memory-operations/Overview.md
```diff
@@ -272,43 +286,24 @@ The instruction has the signature `[i32 i32 i32] -> []`. The parameters are, in
...
+or the target memory.  Bounds checking is performed before any data are written.
+The order of writing is unspecified.
```
rossberg: It's worth noting that this is not currently observable, so does not compromise determinism.

lars-t-hansen:
@rossberg, ptal and make sure i did not undo any of your work when i merged :-)  If it looks OK you can go ahead and merge (I don't think I have privileges).

proposals/bulk-memory-operations/Overview.md
```diff
@@ -171,93 +169,111 @@ Since WebAssembly currently does not allow for multiple memories, the memory
...
+This behavior is changed in the bulk memory proposal:
 
-The behavior of element segment initialization is changed similarly, with the
-difference that elements are copied from element segments into tables, instead
-of bytes being copied from data segments into memories.
+Each active segment is initialized in module-definition order. For
+each segment, if reading the source or writing the destination would
+go out of bounds, then instantiation fails at that point.  Data that
+had already been written for previous (in-bounds) segments stays
+written.
```
aheejin: While for all other instructions the bounds checks are done before copying/initialization and partial copying does not occur, does partial copying occur for active segment initialization?

conrad-watt: Each segment is bounds-checked and initialized individually, in-order. If a segment fails a bounds check, previous segments will stay written, but the oob segment will not do any partial write, and the rest of the segments will not be written (immediate trap).

Initialization can be thought of as iteratively doing init -> drop on the active segment declarations. The bounds checking comes precisely from the behaviour of init. In the MVP, by contrast, all segments are bounds-checked before any are written.

aheejin: Oh, right, partial write does not occur _within_ a segment. Thanks.

PR #136, opened 2020-01-23
Add test ensuring data segment index is unsigned

binji:
Decoding 64 as an signed LEB will produce the value -64 = 4294967232.
This change adds tests to ensure that the segment is decoded as 64.


binji:
This is a silly mistake to make. But I made it, so perhaps we should add it to the test suite. :-)

PR #137, opened 2020-02-05
Fix expected validation errors for `elem.drop`

fitzgen:
The validation for `elem.drop` only checks that the referenced segment is valid within the context, it does not check the existence of tables. Therefore, when asserting that a module is invalid because of a bad segment index in `elem.drop`, we should expect an error message that references an unknown element segment, not a message that references an unknown table.

### DIFF
--- a/interpreter/valid/valid.ml
+++ b/interpreter/valid/valid.ml
@@ -261,7 +261,6 @@ let rec check_instr (c : context) (e : instr) (s : infer_stack_type) : op_type =
     [I32Type; I32Type; I32Type] --> []
 
   | ElemDrop x ->
-    ignore (table c (0l @@ e.at));
     ignore (elem c x);
     [] --> []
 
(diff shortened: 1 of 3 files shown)


binji:
These tests are run in the reference interpreter, where they currently pass. It looks like the issue is the [ElemDrop instruction validation](https://github.com/WebAssembly/bulk-memory-operations/blob/master/interpreter/valid/valid.ml#L264), which checks for the existence of table 0. Do  you mind fixing this (and `data.drop` too, which has the same issue)?

fitzgen:
Okay, I fixed the spec interpreter's validation and also updated the expected error messages for invalid `data.drop`s in `memory_init.wast`.

Let's definitely wait on CI since I did this all in the online github editor :)

fitzgen:
Okay, CI is finally green :)

PR #139, opened 2020-02-26
Test that partially out of bounds segments don't write to a memory/table

### LINKED ISSUE #138: Should partial memory/table initializations after instantiation traps still be visible?
Thinking particularly about these tests:

* https://github.com/WebAssembly/bulk-memory-operations/blob/404b7e7b4276ab3b772a03a1786d5277d6e24836/test/core/linking.wast#L227

* https://github.com/WebAssembly/bulk-memory-operations/blob/404b7e7b4276ab3b772a03a1786d5277d6e24836/test/core/linking.wast#L336

IIRC `memory.init` and `table.init` didn't used to do up front bounds checks, which would explain the behavior prescribed by those tests.

However, the spec now has bounds checks before any mutations for `memory.init` and `table.init`. Since [instantiation](https://webassembly.github.io/bulk-memory-operations/core/exec/modules.html#instantiation) of active table segments and data segments is described in terms of `memory.init` and `table.init`, I would expect the bounds checks to happen before any mutations to the memory and table.

Am I overlooking something?

fitzgen:
Fixes #138

### DIFF
--- a/test/core/linking.wast
+++ b/test/core/linking.wast
@@ -231,11 +231,12 @@
     (table (import "Mt" "tab") 10 funcref)
     (func $f (result i32) (i32.const 0))
     (elem (i32.const 7) $f)
-    (elem (i32.const 12) $f)  ;; out of bounds
+    (elem (i32.const 8) $f $f $f $f $f)  ;; (partially) out of bounds
   )
   "out of bounds"
 )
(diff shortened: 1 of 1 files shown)


fitzgen:
@binji care to merge this PR? thanks!

binji:
Oops, sorry about that!

PR #140, opened 2020-02-28
Test that drop instructions work w/on memory/table

binji:
This was fixed in the interpreter in the last PR, but only fixed the
messages in various assertions. This commit adds a new explicit test for
this.

### DIFF
--- a/test/core/bulk.wast
+++ b/test/core/bulk.wast
@@ -179,7 +179,6 @@
 ;; Test that the data segment index is properly encoded as an unsigned (not
 ;; signed) LEB.
 (module
-  (memory 1)
   ;; 65 data segments. 64 is the smallest positive number that is encoded
   ;; differently as a signed LEB.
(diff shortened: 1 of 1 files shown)


test/core/bulk.wast
```diff
@@ -292,6 +293,9 @@
   (elem funcref)
   (func (elem.drop 64)))
 
+;; No table is required for the elem.drop instruction.
+(module (elem funcref) (func (elem.drop 0)))
```
rossberg: Same here.

binji: Same what?

test/core/bulk.wast
```diff
@@ -193,6 +192,9 @@
   (data "")
   (func (data.drop 64)))
 
+;; No memory is required for the data.drop instruction.
+(module (data "") (func (data.drop 0)))
```
rossberg: Maybe use a non-empty segment, so that the drop isn't a nop.

binji: Done.

PR #141, opened 2020-03-09
Fix variable names in `memory.init` execution

fitzgen:
The values popped from the stack were named `cnt`, `src`, and `dst` but then later referred to as `n`, `s`, and `d` respectively. I've updated all uses to the names given when popped from the stack.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -736,31 +736,31 @@ Memory Instructions
 
 15. Pop the value :math:`\I32.\CONST~dst` from the stack.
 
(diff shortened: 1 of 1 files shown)


binji:
I think @rossberg preferred the single letter names; in particular, ISTR without them you need to bracket the letters (or they'll be interpreted as implicit multiplication). In any case, these are the only places in the file where `cnt`, `dst` and `src` are used.

nomeata:
https://webassembly.github.io/bulk-memory-operations/core/exec/instructions.html#exec-memory-init is currently inconsistent. Looks like someone needs to make a convicing call which names to use.

rossberg:
@nomeata, the spec doc on this proposal is inconsistent in a number of ways, due to its circular dependency with the reference types proposal. This should be rectified in the ref types spec draft, which includes this one (at some point backporting clean-ups became too much effort).

PR #149, opened 2020-05-13
Update `ref.null` encoding after subtyping change

binji:
Issue https://github.com/WebAssembly/reference-types/issues/69 requires
that `ref.null` instructions include a reference type immediate. This
concept isn't present in the bulk-memory proposal, but the encoding is
(in element segment expressions).

This change updates the binary and text format, but not the syntax. This
is OK for now, since the only reference type allowed here is `funcref`.

### DIFF
--- a/document/core/binary/modules.rst
+++ b/document/core/binary/modules.rst
@@ -343,7 +343,7 @@ It decodes into a vector of :ref:`element segments <syntax-elem>` that represent
    \production{element kind} & \Belemkind &::=&
      \hex{00} &\Rightarrow& \FUNCREF \\
    \production{element expression} & \Belemexpr &::=&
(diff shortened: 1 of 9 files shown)


binji:
> I assume this just cherry-picks pieces from the ref-types repo, so rubber-stamp.

I originally was going to do this, but it turns out that there's a lot of new infrastructure that ref-types repo adds that bulk memory doesn't have. So I just wrote this from scratch, assuming that all we really care about is forward-compatibility with the binary and text formats.

rossberg:
Ah, okay. I guess that's fine. The ref-types repo subsumes the bulk one at this point, so hopefully we won't have to do tricky merging of the two in the future.

rossberg:
I did another merge from bulk into ref-types, to catch up with some recent additions to tests and overview.

Issue #153, opened 2020-06-03
Incompatibility with long LEB128 form from the core spec


RReverser:
While implementing bulk memory proposal in my parser, I've noticed regressions in some core spec tests. In particular these two are very representative: https://github.com/WebAssembly/bulk-memory-operations/blob/66d90a7de541a9954af09cb9777f20cc0bd8bbe1/test/core/binary-leb128.wast#L24-L39

Here, a `0` table index (the only one allowed previously) is encoded in a suboptimal form of `0x80 0x00`.

The new layout used by this proposal relies on the table index previously being always zero and encoded as `0x00`, and assigns a different meaning to byte at the same position as `0x80` and expects it to be one of the following values:

![image](https://user-images.githubusercontent.com/557590/83692451-ec628a80-a5eb-11ea-9e0e-9e61604420ae.png)

![image](https://user-images.githubusercontent.com/557590/83692461-f5ebf280-a5eb-11ea-9d10-72599a79538e.png)

Such long form is likely not something used by implementations in the wild, but it might be worth double-checking and either supporting such long form too, or at least making an explicit decision about this backward-incompatible binary representation change and removing (?) corresponding tests.

gahaas:
I saw and fixed this issue already in the reference-type proposal. I guess we should also apply the fix here. I created a PR for that.

RReverser:
@gahaas Note that the test above (for data section) exhibits the same problem, as well as 2 more other tests which previously used long form of zero.

RReverser:
Just looked up, these are the other 2 tests incompatible with new layout (aside from 2 above): https://github.com/WebAssembly/bulk-memory-operations/blob/66d90a7de541a9954af09cb9777f20cc0bd8bbe1/test/core/binary.wast#L118-L136

RReverser:
@binji @gahaas Is this going to be resolved in one way or another before moving to stage 4? Seems dangerous from backward compatibility PoV.

binji:
I doubt the long form is used anywhere (it would only happen if someone did it manually), so the backward compatibility risk seems relatively low. But you're right that it would be good to mention and remove tests.

RReverser:
> it would only happen if someone did it manually

It could in theory also happen if an encoder used a 5-byte-padded LEB128 form to start with (like e.g. LLVM does at initial stage), but then didn't optimise it away in post-processing.

binji:
[...]

PR #154, opened 2020-06-04
[test] Fix table index encoding

gahaas:
The encoding of the table index in binary-leb128.wast is incorrect with the bulk-memory extensions, see https://github.com/WebAssembly/bulk-memory-operations/issues/153. I saw and fixed the issue first in the reference types proposal (see https://github.com/WebAssembly/reference-types/pull/95), but apparently it also exists here.

### DIFF
--- a/test/core/binary-leb128.wast
+++ b/test/core/binary-leb128.wast
@@ -33,9 +33,10 @@
   "\00asm" "\01\00\00\00"
   "\04\04\01"                          ;; Table section with 1 entry
   "\70\00\00"                          ;; no max, minimum 0, funcref
-  "\09\07\01"                          ;; Element section with 1 entry
+  "\09\09\01"                          ;; Element section with 1 entry
+  "\02"                                ;; Element with explicit table index
   "\80\00"                             ;; Table index 0, encoded with 2 bytes
-  "\41\00\0b\00"                       ;; (i32.const 0) with no elements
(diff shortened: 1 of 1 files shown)


test/core/binary-leb128.wast
```diff
@@ -33,9 +33,10 @@
   "\00asm" "\01\00\00\00"
   "\04\04\01"                          ;; Table section with 1 entry
   "\70\00\00"                          ;; no max, minimum 0, funcref
-  "\09\07\01"                          ;; Element section with 1 entry
+  "\09\09\01"                          ;; Element section with 1 entry
```
RReverser, line 36: Just so that we're clear: this is doing the 2nd alternative

> or at least making an explicit decision about this backward-incompatible binary representation change and removing (?) corresponding tests

from my issue description, right?

Because this changes the test itself rather than the spec, it basically acknowledges that new layout is a breaking change, but we think it's fine and won't break anything in the wild?

RReverser: @gahaas @binji Ping. Was there any resolution here?

binji: Since this change was already made in https://github.com/WebAssembly/reference-types/pull/95, it makes sense to do the same here.

RReverser:
...Ping? :)

RReverser:
@gahaas Can you take care of the duplicates mentioned above too?

